PR #10, opened 2021-07-12
PE-50: Describe price calculation as estimation rather than total

fedellen:
There is a natural delay between price estimation and when the total AR taken from the user's wallet. This estimation can be improved in core-js, but when uploading many larger files the delay becomes larger and a volatile AR price could be much different in 10-15 minutes when a large transaction has settled.

This PR simply clarifies that the prompt is an estimation, rather than the total that will be charged.

### DIFF
--- a/src/prompts.ts
+++ b/src/prompts.ts
@@ -474,7 +474,7 @@ const promptForArDriveUpload = async (
 	autoSyncApproval: number
 ): Promise<boolean> => {
 	console.log(
-		'Uploading %s files, %s folders and %s changes (%s) to the Permaweb, totalling %s AR / %s USD',
+		'Uploading %s files, %s folders and %s changes (%s) to the Permaweb, estimated cost: %s AR / %s USD',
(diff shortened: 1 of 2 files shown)


src/prompts.ts
```diff
@@ -474,7 +474,7 @@ const promptForArDriveUpload = async (
 	autoSyncApproval: number
 ): Promise<boolean> => {
 	console.log(
-		'Uploading %s files, %s folders and %s changes (%s) to the Permaweb, totalling %s AR / %s USD',
+		'Uploading %s files, %s folders and %s changes (%s) to the Permaweb, estimated cost: %s AR / %s USD',
```
arielmelendez, line 477: Nit: Looks like a readable English sentence, so consider `... with estimated cost: ...`. Also what's our official company policy on Oxford commas 😆 ?

arielmelendez (approved):
LGTM

jdaev (approved):
LGTM

agsuy (approved):

PR #12, opened 2021-07-28
docs(devSetup): Update README with complete and accurate dev env setup PE-78

fedellen:


jdaev (dismissed):
Looks good 👍🏽

README.md
```diff
@@ -2,38 +2,77 @@
...
 4. Use the correct version of Node, by using: `nvm use`
 
-### We use Yarn 2
+### ArDrive-Core-JS
 
-```
-$ yarn install
-$ ardrive-cli
-```
+You'll need a build of the `ardrive-core-js` library on your local system to run the CLI:
```
agsuy, line 46: This could be added as advice to test a local version of Ardrive Core. e.g. if you want to validate a PR

Otherwise we should be importing a published ardrive-core package. That line on `package.json `was there to ease our dev flow but should never reach prod.

agsuy (changes requested):
Please move the advice on using a local core version to a new section with testing/dev guidelines

agsuy (approved):

PR #14, opened 2021-08-04
chore(version): Use core version `0.4.0` PE-95

fedellen:
This PR locks the CLI to `ardrive-core-js` version `0.4.0` and increments CLI version to `0.3.0` in preparation for releasing Core/CLI as npm packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
 	"dependencies": {
 		"@types/node": "^14.14.32",
 		"@types/uuid": "^8.3.0",
-		"ardrive-core-js": "REPLACE WITH you local ardrive-core",
+		"ardrive-core-js": "0.4.0",
 		"arweave": "1.10.11",
 		"arweave-bundles": "^1.0.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,67 +1,67 @@
...
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/ardriveapp/ardrive-cli.git"
-	},
-	"files": [
-		"lib/**/*"
-	]
-}
+{
+	"name": "ardrive-cli",
```
arielmelendez, line 2: Did your IDE impose some whitespace changes here?

arielmelendez (approved):
Approved with a question about whitespace.

package.json
fedellen: Yes, looks like it fixed the tabbing here from `2` to `4`. Seems that the .json files are skipped by the prettier format command, but when `.editorconfig` is setup my IDE formats .json files correctly

PR #15, opened 2021-08-05
chore(version): Update core version and yarn cache PE-95

fedellen:
This PR sets the target `ardrive-core-js` version to the `latest` tag and commits that version to the `.yarn/cache`


matibat (approved):
Looking good to me. Pretty atomic.

package.json
```diff
@@ -8,7 +8,7 @@
 	"dependencies": {
 		"@types/node": "^14.14.32",
 		"@types/uuid": "^8.3.0",
-		"ardrive-core-js": "0.4.0",
+		"ardrive-core-js": "latest",
```
arielmelendez, line 11: This seems presumptive that we'll always be compatible with the latest version. Could we just peg to the best known version for now?

arielmelendez (changes requested):
Let's peg to a known good version unless we plan to always have CLI move in lock step with latest core version.

arielmelendez (approved):

PR #19, opened 2021-09-07
chore(cli_v2): Cleans argument of sendARToAddress() PE-204

matibat:


src/wallet_new.ts
```diff
@@ -94,10 +94,12 @@ export class WalletDAO {
...
 		fromWallet: Wallet,
 		toAddress: ArweaveAddress,
-		appName = 'ArDrive-Core',
-		appVersion = '1.0',
-		trxType = 'transfer',
-		otherTags?: GQLTagInterface[]
+		[
+			{ value: appName = 'ArDrive-Core' },
+			{ value: appVersion = '1.0' },
+			{ value: trxType = '1.0' },
```
arielmelendez, line 100: default value should be `transfer` rather than `1.0`

matibat: Here: e8cd2d4c8277f5bbb0fc5f22bdd5ffe3a003885e

arielmelendez (approved):

PR #22, opened 2021-09-08
CLI v2:  Removes the signer class

matibat:
… the mutability PE-204

### DIFF
--- a/src/arfsdao.ts
+++ b/src/arfsdao.ts
@@ -8,53 +8,35 @@ export const ArFS_O_11 = '0.11';
 
 export type CipherIV = string;
 
-export type ArFSPrivateObjectData = {
-	cipher: CipherType;
-	cipherIV: CipherIV;
-	driveAuthMode: DriveAuthMode;
-};
-export abstract class ArFSDriveSigner {
-	abstract getPrivacyType(): DrivePrivacy;
-}
-
-export class ArFSPublicDriveSigner extends ArFSDriveSigner {
-	getPrivacyType(): DrivePrivacy {
-		return 'public';
-	}
-}
-
-export class ArFSPrivateDriveSigner extends ArFSDriveSigner {
-	constructor(readonly privacyData: ArFSPrivateObjectData) {
-		super();
-	}
-
-	getPrivacyType(): DrivePrivacy {
-		return 'private';
-	}
-}
-
 export abstract class ArFSDriveMetaDataPrototype {
 	abstract driveName: string;
 	abstract rootFolderId: string;
 	abstract unixTime: number;
 	abstract driveId: string;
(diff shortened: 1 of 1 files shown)


src/arfsdao.ts
```diff
@@ -8,53 +8,35 @@ export const ArFS_O_11 = '0.11';
...
-}
-
 export abstract class ArFSDriveMetaDataPrototype {
 	abstract driveName: string;
 	abstract rootFolderId: string;
 	abstract unixTime: number;
 	abstract driveId: string;
-	abstract signer: ArFSDriveSigner;
+	abstract privacy: DrivePrivacy;
+	abstract cipher: CipherType;
```
arielmelendez, line 17: I think the risk of moving cipher and cipherIV into the base type is that it increases the probability of using the type incorrectly. Not every type of drive needs the fields (i.e. a public drive), so this can create confusion for clients of the class, and, worse, could lead to someone making changes in the subtypes later on that accidentally end up using the fields in a way that is not ArFS compliant. So to reiterate the goals (whether achieved in the current state or not):
• Make types easy to use correctly and hard to use incorrectly
• Let the types express and facilitate the business logic rather than sets of procedures on loosely structured data.

I recognize the helpfulness of typing the implementations of those members in the subclass as "never", but that seems like a bandaid for the door having been opened to a developer error later on.

For what it's worth, I intended to use the signer type to provide for encrypting the data that should be in the transaction as well as the currently presented role of adding the necessary tag data to it. But after looking at Core's code more closely, I've come around to the idea that the cipher-related fields are simply outputs of an encryption process that need to be tracked. But I DO want to make sure that those outputs are tightly coupled to the 'private' privacy type and so I think it's best to keep those pieces of data moving together. Here's the code from core that encrypts the drive metadata:
```
          if (drive.drivePrivacy === 'private') {
			console.log('Creating a new Private Drive (name: %s) on the Permaweb', drive.driveName);
			const driveKey: Buffer = await deriveDriveKey(user.dataProtectionKey, drive.driveId, user.walletPrivateKey);
			const encryptedDriveMetaData: ArFSEncryptedData = await driveEncrypt(
				driveKey,
				Buffer.from(driveMetaDataJSON)
			);
			drive.cipher = encryptedDriveMetaData.cipher;
			drive.cipherIV = encryptedDriveMetaData.cipherIV;
			transaction = await arweave.prepareArFSDriveTransaction(user, encryptedDriveMetaData.data, drive);
		}
```

Let's sync up in the morning to improve upon what we have so far.

arielmelendez: See my updates in 1f20c30. Hopefully they address your concerns about the signer class.

matibat:
We have already take a similar approach. Closing.

PR #26, opened 2021-09-10
Cli v2 upload private file

arielmelendez:


src/arfsdao.ts
```diff
@@ -408,15 +616,6 @@ export class ArFSDAO {
 			// Get the drives transaction ID
 			drive.txId = node.id;
 		});
-
```
arielmelendez, line 411: I think the diff got weird because of function ordering... but is there code missing here for private drives?

fedellen: Grabbing the Parent Folder ID is missing here. I think when we try to create a private folder with only a drive ID, that transaction will be missing the "Parent-Folder-Id" tag

We will need this from the public drive getter:

```ts

		const txData = await this.arweave.transactions.getData(drive.txId, { decode: true });
		const dataString = await Utf8ArrayToStr(txData);
		const dataJSON = await JSON.parse(dataString);

		// Get the drive name and root folder id
		drive.name = dataJSON.name;
		drive.rootFolderId = dataJSON.rootFolderId;
		```

File: src/arfs_prototypes.ts [hunk 2]
```diff
@@ -185,25 +234,25 @@ export class ArFSFileDataPrototype extends ArFSObjectMetadataPrototype {
 	}
 }
 
-export class ArFSPublicFileMetaDataPrototype extends ArFSFileMetaDataPrototype {
-	readonly privacy: DrivePrivacy = 'public';
+export class ArFSPublicFileDataPrototype extends ArFSFileDataPrototype {
+	constructor(readonly objectData: ArFSPublicFileDataTransactionData, readonly contentType: DataContentType) {
+		super();
+	}
+}
```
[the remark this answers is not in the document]
matibat: Ohh it's taken at the constructor. Nevermind

matibat (approved):
The issue of fetching a yet unexisting drive is now fixed. It's looking good.

fedellen (approved):

PR #27, opened 2021-09-10
Cli v2 query builder

fedellen:


matibat (approved):
The only thing I wanna point out is we'll need later to make the `builderQuery()` to allow choosing a different configuration of `edgesFragment` if we want to retrieve only what we need in some specific querys.

For now, I think it works for any cases so it's functional :nerd_face:

src/query.ts
```diff
@@ -0,0 +1,47 @@
...
+			id
+			tags {
+				name
+				value
+			}
+		}
+	}
+`;
+
+// interface ArFSTagInterface extends GQLTagInterface {
```
arielmelendez, line 15: nit: cleanup

src/query.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Builds a GraphQL query which will only return the latest result
+ *
+ * TODO: Add parameters and support for all possible upcoming GQL queries
+ *
+ * @example
+ * const query = buildQuery([{ name: 'Folder-Id', value: folderId }]);
+ */
+export function buildQuery(tags: GQLTagInterface[]): { query: string } {
```
arielmelendez, line 27: Can we add a type for the return data?

src/query.ts
```diff
@@ -12,9 +12,7 @@ const edgesFragment = `
 	}
 `;
 
-// interface ArFSTagInterface extends GQLTagInterface {
-// 	name: 'Drive-Id' | 'Folder-Id';
-// }
+type GQLQuery = { query: string };
```
matibat, line 15: We might end using this type on the tests
```suggestion
export type GQLQuery = { query: string };
```

arielmelendez (approved):

PR #29, opened 2021-09-10
CLI v2: get-drive commander setup

matibat:
…et-drive PE-204


fedellen:
This branch was merged directly into cli_v2_get_drives_arfs during development

PR #31, opened 2021-09-13
chore(testing stack): Add testing stack to CLI PE-204

fedellen:
This PR adds the testing stack from `ardrive-core-js` to the CLI repo


matibat (approved):
It works just as core does :+1:

PR #34, opened 2021-09-15
PE-315: Add community oracles and send community tip on file data uploads

fedellen:
This PR sends the ArDrive community tip to a selected token holder.

Architecture for community oracles:

`ArDrive` -> `CommunityOracle` -> `ContractOracle` -> `ContractReader` (Verto / Smartweave / Redstone)

The Community oracle handles selecting token holders and determining the community tip value in winston. 

The Contract oracle handles retrieving the community contract from a Contract reader, and will fallback to another reader if the first fails. It currently uses the Verto cache first, then will fallback to a smartweave-js implementation, and then if both fail it will use redstone-smartweave.

The contract oracle also handles gathering ArDrive community contract specific data from the contract, such as the community `fee` value.

The Contract readers only handle fetching a smart contract for each given reader implementation


fedellen:
This will be marked as a draft temporarily, the branch will not compile until the version is bumped to `ardrive-core-js` v0.5.1

matibat (commented):
The code looks clean and it compiles if you use the right core version. Let's wait until core is released to npm.

src/ardrive.ts
```diff
@@ -11,9 +17,9 @@ export interface ArFSEntityData {
 
 // TODO: Is this really in the ArFS domain?
 export interface ArFSTipData {
-	type: ArFSTipType;
 	txId: TransactionID; // TODO: make a type that checks lengths
-	winston: number; // TODO: make a type that checks validity
+	tokenHolder: ArweaveAddress;
```
fedellen, line 21: recipient makes more sense here

src/community/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,77 @@
...
+ * and selecting the PST token holder for tip distribution
+ * */
+export class ArDriveCommunityOracle implements CommunityOracle {
+	constructor(readonly arweave: Arweave) {
+		this.contractOracle = new ArDriveContractOracle(arweave);
+	}
+
+	private readonly contractOracle: ContractOracle;
+
+	/** Given an AR data cost, returns a calculated ArDrive community tip amount in AR */
```
fedellen, line 25: Add TODO for Big Int on ALL Winston types

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	/**
+	 * Array of contract readers to use as fall back if one fails
+	 * Uses contract reader at index 0 first then descends down the list
+	 */
+	private contractReaders: ContractReader[] = [
+		new VertoContractReader(),
+		new SmartweaveContractReader(this.arweave),
+		new RedstoneContractReader(this.arweave)
```
fedellen, line 39: Remove restone while still in alpha. One fallback is enough

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,143 @@
...
+	/**
+	 * Reads the ArDrive Community Contract with the current contract reader
+	 *
+	 * @remarks Will fallback to other contract readers when one fails
+	 */
+	async readContract(txId: string): Promise<unknown> {
+		let contract: unknown;
+
+		try {
+			contract = await this.contractReaders[this.currentContractReader].readContract(txId);
```
fedellen, line 68: Comment here to explain that its the contract reader's readContract implementation

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,143 @@
...
+			return contract;
+		} catch (error) {
+			this.readContractAttempts++;
```
fedellen, line 71: Add 2 retries for Verto before falling back to smartweave

src/community/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,77 @@
...
+	async getCommunityWinstonTip(winstonCost: Winston): Promise<Winston> {
+		const communityTipValue = await this.contractOracle.getTipSettingFromContractSettings();
+		const arDriveCommunityTip = +winstonCost * (communityTipValue / 100);
+		return Math.round(Math.max(arDriveCommunityTip, minArDriveCommunityWinstonTip)).toString();
+	}
+
+	/** Gets a random ArDrive token holder based off their weight (amount of tokens they hold)  */
+	async selectTokenHolder(): Promise<ArweaveAddress> {
```
fedellen, line 33: Add TODO for unit tests on important functions -- like `selectTokenHolder`

src/community/ardrive_contract_oracle.ts
fedellen: Now retries twice per contract reader:

```shell
➞  yarn ardrive upload-file -f 38123551-00fc-4e2e-9576-3279bc406321 -w CLI_TEST_arweave-keyfile-g1hzNXVbh2M6LMQSUYp7HgkgxdadYqYEfw-HAajlms0.json -l crown.txt        
Contract could not fetched: FetchError: request to https://v2.cache.verto.exchange/-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ failed, reason: connect ECONNREFUSED 127.0.0.1:443
Retrying with current contract reader..
Contract could not fetched: FetchError: request to https://v2.cache.verto.exchange/-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ failed, reason: connect ECONNREFUSED 127.0.0.1:443
Retrying with current contract reader..
Contract could not fetched: FetchError: request to https://v2.cache.verto.exchange/-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ failed, reason: connect ECONNREFUSED 127.0.0.1:443
Falling back to next contract reader..
{
    "created": [
        {
            "type": "file",
            "metadataTxId": "kUp8LwM1XnYeDY9x0BIjs7dZDhJcsYRxpqUzpSQ4PwY",
            "dataTxId": "kSfYy3z0Kou12unVFDVGW5weO9mIbU_PpD2Weg2xDj8",
            "entityId": "d60551a5-f0e7-448b-9f05-098fa658f22c"
        }
    ],
    "tips": [
        {
            "txId": "s-sBRVQK2kpPhZazSvDhVXix1jLDr0VL3GzZtF2A18Q",
            "recipient": "Zznp65qgTIm2QBMjjoEaHKOmQrpTu0tfOcdbkm_qoL4",
            "winston": "10000000"
        }
    ],
    "fees": {
        "kUp8LwM1XnYeDY9x0BIjs7dZDhJcsYRxpqUzpSQ4PwY": 1533072,
        "kSfYy3z0Kou12unVFDVGW5weO9mIbU_PpD2Weg2xDj8": 1516656
    }
}
```

src/ardrive.ts
```diff
@@ -39,23 +45,64 @@ export class ArDriveAnonymous extends ArDriveType {
 	}
 }
 
+// TODO: ArDrive should accept App-Name and App-Version tags from constructor?
```
arielmelendez, line 48: Good idea

src/ardrive.ts
```diff
@@ -39,23 +45,64 @@ export class ArDriveAnonymous extends ArDriveType {
...
+const commTipMetaTags: GQLTagInterface[] = [
+	{ name: 'App-Name', value: 'ArDrive-CLI' },
+	{ name: 'App-Version', value: '2.0' },
+	{ name: 'Trx-Type', value: 'Community-Tip' }
```
arielmelendez, line 52: Is this a new tag that we'll be adding to community tips? Here's how they're constructed on web;
```
feeTx = await _arweave.client.transactions.prepare(
        Transaction(
          target: await _pst.getWeightedPstHolder(),
          quantity: pstFee,
        ),
        profile.wallet,
      )
        ..addApplicationTags()
        ..addTag('Type', 'fee')
        ..addTag(TipType.tagName, TipType.dataUpload);
      await feeTx!.sign(profile.wallet);
```

src/wallet_new.ts
```diff
@@ -4,12 +4,15 @@ import * as crypto from 'crypto';
 import jwkToPem, { JWK } from 'jwk-to-pem';
 import Arweave from 'arweave';
 import * as mnemonicKeys from 'arweave-mnemonic-keys';
+import { Bytes } from './ardrive';
```
arielmelendez, line 7: Note for the future: we should push some of these typealiases down into a separate file to avoid circular dependencies among the source files.

src/wallet_new.ts
```diff
@@ -4,12 +4,15 @@ import * as crypto from 'crypto';
...
 
 type PublicKey = string;
 export type ArweaveAddress = string;
 type SeedPhrase = string;
 type TransactionID = string;
-type Winston = string;
+
+/** TODO: Use big int library on Winston types */
```
arielmelendez, line 14: What use cases do you have in mind?

src/wallet_new.ts
```diff
@@ -78,6 +81,15 @@ export class WalletDAO {
 		return Promise.resolve(+(await this.arweave.wallets.getBalance(address)));
 	}
 
+	async confirmBalance(wallet: Wallet, winstonPrice: Winston): Promise<boolean> {
```
arielmelendez, line 84: for readability, consider `async walletHasBalance`. Usage:
```
if (!this.walletDao.walletHasBalance(this.wallet, totalWinstonPrice)) {
    throw new Error('Not enough AR for file upload..');
}
```

src/wallet_new.ts
```diff
@@ -78,6 +81,15 @@ export class WalletDAO {
...
+		const walletBalance = await this.getWalletWinstonBalance(wallet);
+		return +walletBalance > +winstonPrice;
+	}
+
+	async getWinstonPriceForBytes(bytes: Bytes): Promise<Winston> {
```
arielmelendez, line 89: Might want to add a TODO here to use the price estimator utility when once we reintegrate it with core.

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+	 */
+	async readContract(txId: string): Promise<unknown> {
+		let contract: unknown;
+
+		try {
+			// Get contract with current contract reader's readContract implementation
+			contract = await this.contractReaders[this.currentContractReader].readContract(txId);
+			return contract;
+		} catch (error) {
+			console.error(`Contract could not fetched: ${error}`);
```
arielmelendez, line 74: nit: grammar

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Oracle class responsible for retrieving the correct data fields from
+ * the ArDrive Community Contract. This class can utilize several different
+ * contract readers and will fallback to other readers if one fails
+ *
+ * @remarks Will begin fetching data from default contract reader on construction
+ */
+export class ArDriveContractOracle implements ContractOracle {
+	constructor(private readonly arweave: Arweave) {
```
arielmelendez, line 31: consider a `skipSetup` flag for unit testing purposes.

arielmelendez: Might add more complexity than we want for this ticket right now so a TODO is fine.

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+
+// ArDrive Profit Sharing Community Smart Contract
+export const communityTxId = '-8A6RexFkpfWwuyVO98wzSFZh0d6VJuI-buTJvlwOJQ';
+
+const maxReadContractAttempts = 3;
+const initialContractReader = 0;
+const initialContractAttempts = 0;
+
+export interface CommunityContractData {
+	votes: [Record<string, unknown>];
```
arielmelendez, line 17: Is this the best we can do for typing the contract data?

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * @remarks Will begin fetching data from default contract reader on construction
+ */
+export class ArDriveContractOracle implements ContractOracle {
+	constructor(private readonly arweave: Arweave) {
+		// Get contract data upon construction
+		this.getCommunityContract();
+	}
+
+	/**
```
arielmelendez, line 36: Nit: hoist these above the constructor

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+	 * Array of contract readers to use as fall back if one fails
+	 * Uses contract reader at index 0 first then descends down the list
+	 */
+	private contractReaders: ContractReader[] = [
```
arielmelendez, line 40: Consider making these injectable so that we can make each of the possible contract readers configurable for the CLI, e.g. provide a flag for the preferred contract reader

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+		new VertoContractReader(),
+		new SmartweaveContractReader(this.arweave)
+		// new RedstoneContractReader(this.arweave)
+	];
+	private currentContractReader = initialContractReader;
```
arielmelendez, line 45: Rather than do this statefully, consider chaining the requests functionally with promises.

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+				}
+
+				// Else fallback to next reader
+				this.fallbackToNextContractReader();
+			} else {
+				console.log('Retrying with current contract reader..');
+			}
+
+			// Retry read contract if no error was thrown
+			return this.readContract(txId);
```
arielmelendez, line 94: Recursive approach is risky and invites anyone who makes changes here to create an infinite loop and blow up the stack. Consider an iterative implementation.

src/community/ardrive_contract_oracle.ts
```diff
@@ -0,0 +1,155 @@
...
+		if (this.contractPromise) {
+			return this.contractPromise;
+		}
+
+		// Begin new contract read; cast result to known ArDrive Community Contract type
+		this.contractPromise = this.readContract(communityTxId) as Promise<CommunityContractData>;
+
+		this.communityContract = await this.contractPromise;
+
+		if (!this.communityContract.settings || !this.communityContract.balances) {
```
arielmelendez, line 119: Here's where I'd like for us to just be able to rely on type casting at runtime to throw the error for us rather than manual inspection of the data.

src/community/redstone_contract_oracle.ts
```diff
@@ -0,0 +1,19 @@
+// import Arweave from 'arweave';
```
arielmelendez, line 1: Might want to leave a comment here about why this exists and is commented out.

src/community/contract_oracle.ts
```diff
@@ -0,0 +1,11 @@
+import { CommunityContractData } from './ardrive_contract_oracle';
+
+/** An oracle interface responsible for reading contracts and retrieving the ArDrive Community Contract */
+export interface ContractOracle extends ContractReader {
+	getCommunityContract(): Promise<CommunityContractData>;
+	getTipSettingFromContractSettings(): Promise<number>;
+}
+
+export interface ContractReader {
+	readContract(txId: string): Promise<unknown>;
```
arielmelendez, line 10: Nit: use alias types for `string` and `number` in the file to held readability. We can wait until we unify the price estimator code and this code in core.

src/community/ardrive_community_oracle.ts
```diff
@@ -0,0 +1,87 @@
...
+
+	private readonly contractOracle: ContractOracle;
+
+	/**
+	 * Given an AR data cost, returns a calculated ArDrive community tip amount in AR
+	 *
+	 * TODO: Use big int library on Winston types
+	 */
+	async getCommunityWinstonTip(winstonCost: Winston): Promise<Winston> {
+		const communityTipValue = await this.contractOracle.getTipSettingFromContractSettings();
```
arielmelendez, line 33: Nit: `Value` isn't that useful for comprehension in this variable name. Consider `communityTipPercentage`. Also double check that it's not already a proportion vs. a percentage.

src/wallet_new.ts
arielmelendez: Also I'm not sure this should live in the WalletDAO. But I don't think there's a better spot for it in the hierarchy at the moment.

src/ardrive.ts
```diff
@@ -39,23 +45,64 @@ export class ArDriveAnonymous extends ArDriveType {
...
+
+		return { txId: communityTipResult.trxID, recipient: tokenHolder, winston: communityTipResult.winston };
+	}
+
 	async uploadPublicFile(
 		parentFolderId: FolderID,
 		filePath: string,
 		destinationFileName?: string
 	): Promise<ArFSResult> {
+		const winstonPrice = await this.walletDao.getWinstonPriceForBytes(this.getFileSize(filePath));
```
arielmelendez, line 88: You shouldn't need to do this. You can get the cost of the data transaction directly from the trx object itself, i.e.:
```
dataTrx.reward
```

src/ardrive.ts
```diff
@@ -80,28 +126,39 @@ export class ArDrive extends ArDriveAnonymous {
...
+
 		return Promise.resolve({
 			created: [
 				{
 					type: 'file',
 					metadataTxId: metaDataTrx.id,
 					dataTxId: dataTrx.id,
-					entityId: fileId
+					entityId: fileId,
+					key: ''
```
arielmelendez, line 155: We need to implement returning the file keys here.

src/ardrive.ts
```diff
@@ -80,28 +126,39 @@ export class ArDrive extends ArDriveAnonymous {
...
 				}
 			],
-			tips: [],
+			tips: [communityTipResult],
 			fees: {
 				[metaDataTrx.id]: +metaDataTrx.reward,
 				[dataTrx.id]: +dataTrx.reward
-				// qGr1BIVWQwdPMuQxJ9MmwMM8CBmZTIj9powGxJSZyi0: 344523
```
arielmelendez, line 104: I believe we should be adding the network fee for the tip trx here.

src/ardrive.ts
```diff
@@ -80,28 +126,39 @@ export class ArDrive extends ArDriveAnonymous {
...
 	): Promise<ArFSResult> {
+		const winstonPrice = await this.walletDao.getWinstonPriceForBytes(this.getFileSize(filePath));
+		const communityWinstonTip = await this.communityOracle.getCommunityWinstonTip(winstonPrice);
+		const totalWinstonPrice = (+winstonPrice + +communityWinstonTip).toString();
+
+		if (!this.walletDao.confirmBalance(this.wallet, totalWinstonPrice)) {
+			throw new Error('Not enough AR for file upload..');
+		}
+
+		// TODO: Add interactive confirmation of AR price estimation
```
arielmelendez, line 137: When?

src/ardrive.ts
```diff
@@ -65,11 +112,10 @@ export class ArDrive extends ArDriveAnonymous {
 					entityId: fileId
 				}
 			],
-			tips: [],
+			tips: [communityTipResult],
 			fees: {
 				[metaDataTrx.id]: +metaDataTrx.reward,
 				[dataTrx.id]: +dataTrx.reward
-				// qGr1BIVWQwdPMuQxJ9MmwMM8CBmZTIj9powGxJSZyi0: 344523
```
arielmelendez, line 72: I believe we should be adding the network fee for the tip trx here.

src/ardrive.ts
```diff
@@ -39,23 +45,64 @@ export class ArDriveAnonymous extends ArDriveType {
...
+		const tokenHolder: ArweaveAddress = await this.communityOracle.selectTokenHolder();
+
+		const communityTipResult = await this.walletDao.sendARToAddress(
+			winstonToAr(+communityWinstonTip),
+			this.wallet,
+			tokenHolder,
+			commTipMetaTags
+		);
+
+		return { txId: communityTipResult.trxID, recipient: tokenHolder, winston: communityTipResult.winston };
```
arielmelendez, line 80: Let's update the ArFSTipData type to send back the entire transaction so that we can extract the network fee out for the ArDrive layer's return data (i.e. `trx.reward`).

arielmelendez: If we do that, we can probably get rid of all the other props here since it's all in the trx.

arielmelendez (changes requested):
Left a few requests and handful of nits, the latter of which are optional.

src/wallet_new.ts
fedellen: Agustin suggested we take this route to preserve accuracy of larger transactions. 

Arweave.js uses `import { BigNumber } from "bignumber.js";` in all of their `ar` class functions, for example.

src/ardrive.ts
fedellen: The idea here is that we should check the price plus the community and confirm a wallet's balance before creating the transaction. When we create the transaction, we will inevitably be writing file's Buffer to it which the arweave library will encode into base64. 

If the wallet doesn't have enough balance for both the transaction AND the community tip, we shouldn't be initiating something so resource intensive.

Creating the transaction in ArFSDao, handing the transaction back up to check the balance, then sending back down to another function that will sign/send the transaction and is one solution. But that solution would not be performant with more than one file upload at a time -- because we'd be holding on to all of these `dataTrx` types in memory.

src/ardrive.ts
fedellen: Its a phase 1.5 thing once we have an interactivity pattern established. The current CLI will ask the user (can be set auto-accept) to confirm a price estimation before sending the transactions.

src/ardrive.ts
```diff
@@ -39,23 +44,81 @@ export class ArDriveAnonymous extends ArDriveType {
...
+	) {
 		super(arFsDao);
 	}
 
+	// TODO: FS shouldn't be reading the files more than once and doesn't belong in this class
+	getFileSize(filePath: string): Bytes {
+		return fs.statSync(filePath).size;
+	}
+
+	async prepareCommunityTipTrx(communityWinstonTip: Winston): Promise<Transaction> {
```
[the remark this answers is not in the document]
arielmelendez: I think the answer is just to have the price estimator code available here...

src/ardrive.ts
arielmelendez: You're right about performance and the changes you made to try and do that ended up leaking the Arweave library's types out into the ArDrive layer which I don't think we want. We need the price estimator classes.

src/ardrive.ts
arielmelendez: Don't forget about this!

src/ardrive.ts
arielmelendez: I realize now that this comment's suggestion would have broken the architecture and leaked Arweave types into the upper layers.

arielmelendez:
The nit and update resolutions look good EXCEPT for the big change to use Arweave types out in the ArDrive layer. For better or worse, I think the solution might be to bring in the price estimator classes for usage in the ardrive layer.

package.json
```diff
@@ -9,9 +9,10 @@
 	"types": "./lib/index.d.ts",
 	"dependencies": {
 		"@types/node": "^14.14.32",
+		"@types/regression": "^2.0.2",
 		"@types/uuid": "^8.3.0",
-		"ardrive-core-js": "0.5.0",
-		"arweave": "1.10.11",
+		"ardrive-core-js": "0.5.1",
+		"arweave": "^1.10.16",
```
agsuy, line 15: Was this intended? Is arweave using semver?
We moved from a fixed version to any update but a major version...

fedellen: It was intended. The issue I was running into was a package conflict between  the `ardrive-core-js` dependency of `arweave` and the local CLI dependency of `arweave`.

To make sure they didn't conflict again, I specifically copied the version from the core commit below. Core was previously already using the `^` version syntax:

```diff
-		"arweave": "^1.10.11",
+		"arweave": "^1.10.16",
```

 https://github.com/ardriveapp/ardrive-core-js/pull/73/commits/c2b4faa638e7da0aa5fb35eac391f03135769639#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

arielmelendez (approved):

PR #38, opened 2021-09-17
Read-only DAO

matibat:
Adds an implementation of the read-only versions of the DAOs we had.
Uses list-drive command as example for testing


src/arfsdao.ts
```diff
@@ -73,9 +72,110 @@ export interface ArFSCreatePrivateDriveResult extends ArFSCreateDriveResult {
 	driveKey: DriveKey;
 }
 
-export class ArFSDAO {
+/**
+ * contains all queries that doesn't need for a wallet
+ */
+export class ArFSReadOnlyDAO {
```
arielmelendez, line 78: I think it's a little tricky to break up ArFS and DAO in the name like this. I'd follow the naming pattern you used above with ArDriveReadOnly, i.e. ArFSDAOReadOnly

arielmelendez (approved):

src/arfsdao.ts
matibat: Renamed at 5feaea5

arielmelendez (approved):

fedellen (approved):

PR #41, opened 2021-09-23
Reverse arweave leakage

arielmelendez:


src/ardrive.ts
```diff
@@ -44,19 +48,15 @@ export class ArDriveAnonymous extends ArDriveType {
...
-];
-
 export class ArDrive extends ArDriveAnonymous {
 	constructor(
 		private readonly wallet: Wallet,
 		private readonly walletDao: WalletDAO,
 		protected readonly arFsDao: ArFSDAO,
-		private readonly communityOracle: CommunityOracle
+		private readonly communityOracle: CommunityOracle,
+		private readonly appName: string,
```
fedellen, line 57: We should also remember to feed these down into arfsDao. Currently all transactions are still being tagged with 

```
        {
          "name": "App-Name",
          "value": "ArDrive-Core"
        },
        {
          "name": "App-Version",
          "value": "1.0"
        },
```

fedellen (approved):
Looks good! 

I'm still a little concerned that we're now effectively sending those 3 requests to arweave.net for instead of 1 for single file uploads. But, it does make sense for bulk uploads coming up.

src/ardrive.ts
```diff
@@ -126,13 +139,26 @@ export class ArDrive extends ArDriveAnonymous {
 		});
 	}
 
+	async encryptedFileSize(filePath: string, drivePassword: string, driveId: string, fileId: string): Promise<number> {
```
fedellen, line 142: This does seem rather intense to do this here for a price and then throw away the buffer, especially in a bulk upload sense.

I wonder what the results of this function are when graphed out. Is there a standard amount of bytes added on, or does it depend on the file size?

And why do we accept a driveId and fileId here instead of just defining the fake IDs here, do different characters of the same uuid length produce different encrypted sizes?

fedellen (approved):

src/ardrive.ts
arielmelendez: All valid. Am going to look into AES GCM block sizes.

arielmelendez: There's hope: https://stackoverflow.com/questions/3283787/size-of-data-after-aes-cbc-and-aes-ecb-encryption

Note that we need to sort out the padding sizes.

PR #44, opened 2021-09-24
PE-326: Bulk file upload

fedellen:
This PR implements bulk file upload when a CLI user supplies a folder as the `--local-file`  parameter on `upload-file`. 

The command will now detect whether the filePath is a file or a folder. If it is a folder, it will create a folder and upload all files/folders within that folder recursively.

The GQL call for driveId and confirming that the drive exists has been lifted up into the ArDrive layer so that we don't make unnecessary requests for each file. The community tip transaction will calculate the total bytes for all of the files inside the folder, and send one calculated tip to a selected token holder based on that total value.

The above performance improvements have not yet been extended  to the CSV implementation that only lives in the commander layer. CSV bulk uploads should still work mostly the same as before, making 2 GQL calls per file. Alternatively, the filePaths in the CSV will now also accept folder paths and will upload the contents in them just as above.


src/ardrive.ts
```diff
@@ -164,38 +179,196 @@ export class ArDrive extends ArDriveAnonymous {
...
+			}
+		}
+
+		const parentFolderData = new ArFSPublicFolderTransactionData(
+			parentFolderName ?? wrappedFolder.getBaseFileName()
+		);
+
+		let communityWinstonTip = 0;
+
+		if (isParentFolder) {
```
fedellen, line 238: Lift this into createPublicFolderAndUploadChildren now that it exists. This will make it easier to confirm the price later, and avoid passing the tip around needlessly

src/arfsdao.ts
```diff
@@ -40,6 +38,9 @@ import {
 	ArFSPublicFolderTransactionData
 } from './arfs_trx_data_types';
 import { buildQuery } from './query';
+import { FsFile } from './fsFile';
+
+export const ArFS_O_11 = '0.11';
```
fedellen, line 43: This should no longer live in this file, looks like its leftover from a merge conflict

src/arfsdao.ts
```diff
@@ -86,21 +87,32 @@ export interface ArFSUploadPrivateFileResult extends ArFSUploadFileResult {
...
-	folderData: ArFSPublicFolderTransactionData;
+export interface CreateFolderSettings {
 	driveId: DriveID;
 	rewardSettings: RewardSettings;
 	parentFolderId?: FolderID;
 	syncParentFolderId?: boolean;
 }
 
+export interface CreatePublicFolderSettings extends CreateFolderSettings {
+	folderData: ArFSPublicFolderTransactionData;
```
fedellen, line 108: CreatePublicFolderSettings seems redundant here. The base interface can probably hold folderData

src/arfsdao.ts
```diff
@@ -490,10 +544,12 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		const unixTime = Math.round(Date.now() / 1000);
 
 		// Gather file information
-		const fileStats = fs.statSync(filePath);
-		const fileData = fs.readFileSync(filePath);
-		const dataContentType = extToMime(filePath);
-		const lastModifiedDateMS = Math.floor(fileStats.mtimeMs);
+		const fileStats = wrappedFile.fileStats;
```
fedellen, line 547: These all use the wrappedFile class, consider a single gather file information method on the file wrapper for both public and private methods to share

src/arfsdao.ts
fedellen: I was mistaken, the folderData declaration is meant to split the Public from Private types

src/fsFile.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/** Type-guard function to determine if returned class is a File or Folder */
+export function isFolder(fileOrFolder: FsFile | FsFolder): fileOrFolder is FsFolder {
+	return Object.keys(fileOrFolder).includes('files') || Object.keys(fileOrFolder).includes('folders');
+}
+
+export class FsFile {
+	constructor(public readonly filePath: FilePath, public readonly fileStats: fs.Stats) {
+		if (this.fileStats.size >= 2147483647) {
```
fedellen, line 50: Use a const declaration here

src/arfsdao.ts
```diff
@@ -332,6 +399,7 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		// Generate a new drive ID  for the new drive
 		const driveId = uuidv4();
 
+		// TODO: Re-use CreatePrivateFolder function here
```
fedellen, line 402: Add which ticket to do this in, or simply implement it here

src/ardrive.ts
```diff
@@ -447,14 +898,15 @@ export class ArDrive extends ArDriveAnonymous {
 
 	// Provides for stubbing metadata during cost estimations since the data trx ID won't yet be known
 	private stubPublicFileMetadata(
-		filePath: string,
+		wrappedFile: FsFile,
 		destinationFileName?: string
 	): ArFSPublicFileMetadataTransactionData {
-		const fileStats = fs.statSync(filePath);
-		const dataContentType = extToMime(filePath);
+		const fileStats = wrappedFile.fileStats;
```
fedellen, line 904: Use gatherFileInfo method in these stubs

src/fsFile.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+export class FsFile {
+	constructor(public readonly filePath: FilePath, public readonly fileStats: fs.Stats) {
+		if (this.fileStats.size >= maxFileSize) {
+			throw new Error(`Files greater than "${maxFileSize}" bytes are not yet supported!`);
+		}
+	}
+
+	baseCosts?: BulkFileBaseCosts;
```
matibat, line 62: So we need these as optional because we are feeding the value with recursion, there's no easy & clean way to put it in the constructor, right?
That's fine as long as it's hidden behind the getter

matibat (commented):
It's looking good! Left some comments.

Are there changes in `cli_v2` missing in `cli_v2_list_folder` being used in this PR? If not, we can then change the base to `cli_v2_list_folder` to clean the diff

src/fsFile.ts
fedellen: To put these base cost logic inside of the constructor, we'd mainly need to be able access to the price estimator in the layer that we would wrap the file. 

In the code's current state, I've lifted the file wrapping out into the commander layer so we can determine whether we're performing a bulk upload or a single file upload before proceeding into the ArDrive layer.

Making it optional and hiding it behind the getter was my quick solution as I fed the values through the recursive estimation method.

src/fsFile.ts
```diff
@@ -0,0 +1,145 @@
...
+		return basename(this.filePath);
+	}
+
+	/** Computes the size of a private file encrypted with AES256-GCM */
+	public encryptedDataSize(): number {
+		return (this.fileStats.size / 16 + 1) * 16;
+	}
+}
+
+export class FsFolder {
```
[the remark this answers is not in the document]
fedellen: > we can then change the base to cli_v2_list_folder to clean the diff

Good catch, thats what I intended when I merged in the `list_folder` branch. Changed the base branch after I saw your comment on friday 😀

> Maybe rename to 'ArFSFolderToUpload' and 'ArFSFileToUpload'? These classes are helpers for computing & predicting values from the files at the local disk

Added f0910c8

src/commands/upload_file.ts
```diff
@@ -86,7 +92,10 @@ new CLICommand({
 
 			const wallet = readJWKFile(options.walletFile);
 			const priceEstimator: ARDataPriceEstimator = (() => {
-				if (filesToUpload.length > ARDataPriceRegressionEstimator.sampleByteVolumes.length) {
+				if (
+					filesToUpload.length > ARDataPriceRegressionEstimator.sampleByteVolumes.length ||
```
arielmelendez, line 96: Now that we use the estimator for settings rewards on tips and metadata as well, we can always use it when there's at least 1 file since 1 file is 3 requests for pricing data anyway (file, meta, tip).

src/commands/upload_file.ts
```diff
@@ -99,27 +108,42 @@ new CLICommand({
...
-							);
+							if (isFolder(wrappedEntity)) {
+								return arDrive.createPrivateFolderAndUploadChildren(
+									parentFolderId,
+									wrappedEntity,
+									driveKey,
+									destinationFileName
+								);
+							} else {
+								wrappedEntity;
```
arielmelendez, line 129: Straggler?

arielmelendez:
Can we hoist total cost estimation and wallet balance assertion into a bulk spot in the ardrive layer? Given a list of files and folders, we should be able to "stat" them all and do a bulk cost computation. I suppose we can do it in a followup ticket since the only feature that really facilitates is making sure there's enough AR in the wallet before you begin.

src/ardrive.ts
```diff
@@ -213,38 +244,180 @@ export class ArDrive extends ArDriveAnonymous {
...
+	): Promise<ArFSResult> {
+		const driveId = await this.getDriveIdAndAssertDrive(parentFolderId);
+
+		const parentFolderData = new ArFSPublicFolderTransactionData(
+			parentFolderName ?? wrappedFolder.getBaseFileName()
+		);
+
+		// Estimate and assert the cost of the entire bulk upload
+		// This will assign the calculated base costs to each wrapped file and folder
+		const bulkEstimation = await this.estimateAndAssertCostOfBulkUpload(wrappedFolder, 'public', parentFolderData);
```
fedellen, line 267: > Can we hoist total cost estimation and wallet balance assertion into a bulk spot in the ardrive layer?

@arielmelendez 

This is already the case in this PR. The estimation is used here and implemented below in the `estimateAndAssertCostOfBulkUpload` method

src/commands/upload_file.ts
fedellen: Nice hiding spot, cleaned up in this commit a55726c

src/ardrive.ts
```diff
@@ -397,6 +748,80 @@ export class ArDrive extends ArDriveAnonymous {
...
+		folderToUpload: ArFSFolderToUpload,
+		drivePrivacy: DrivePrivacy,
+		parentFolderMetaData?: ArFSObjectTransactionData
+	): Promise<{ totalPrice: Winston; communityWinstonTip: Winston }> {
+		// parentFolderMetaData will only exist if this folder is the parent folder
+		// Recursing children folders will not have meta data assigned
+		const isParentFolder: boolean = parentFolderMetaData !== undefined;
+
+		const metaDataBaseReward = await this.priceEstimator.getBaseWinstonPriceForByteCount(
+			(parentFolderMetaData ?? new ArFSPublicFolderTransactionData(folderToUpload.getBaseFileName())).sizeOf()
```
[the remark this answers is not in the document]
arielmelendez: (The dangers of Stringly typing...)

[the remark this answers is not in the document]
arielmelendez: 3ecc9cd

src/arfs_file_wrapper.ts
```diff
@@ -0,0 +1,145 @@
...
+ */
+const maxFileSize: Bytes = 2147483646;
+
+export interface FileInfo {
+	dataContentType: DataContentType;
+	lastModifiedDateMS: number;
+	fileSize: Bytes;
+}
+
+/**
```
arielmelendez, line 24: Can we clean this up?

src/arfs_file_wrapper.ts
```diff
@@ -0,0 +1,145 @@
...
+	if (entityStats.isDirectory()) {
+		return new ArFSFolderToUpload(fileOrFolderPath, entityStats);
+	}
+
+	return new ArFSFileToUpload(fileOrFolderPath, entityStats);
+}
+
+/** Type-guard function to determine if returned class is a File or Folder */
+export function isFolder(fileOrFolder: ArFSFileToUpload | ArFSFolderToUpload): fileOrFolder is ArFSFolderToUpload {
+	return Object.keys(fileOrFolder).includes('files') || Object.keys(fileOrFolder).includes('folders');
```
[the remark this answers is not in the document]
fedellen: Hmm, no reason. I guess I'm just used to creating  typeguards for non-classes.

That would be much simpler given that we're working with classes here. 

 Added: 6a7aa8e

PR #46, opened 2021-09-26
PE-448: fee boost

arielmelendez:
Adds precise fee estimation to all existing writing commands and the ability to boost miner rewards by a multiple for all commands that send transactions. NOTE: the merge destination branch of `cli_v2_smartweave`. We'll want to merge to the `cli_v2` branch once the former has been merged there.

### DIFF
--- a/src/arfs_trx_data_types.ts
+++ b/src/arfs_trx_data_types.ts
@@ -17,6 +17,7 @@ export interface ArFSObjectTransactionData {
 
 export abstract class ArFSDriveTransactionData implements ArFSObjectTransactionData {
 	abstract asTransactionData(): string | Buffer;
+	// TODO: Share repeated sizeOf() function to all classes
 	sizeOf(): number {
 		return this.asTransactionData().length;
 	}
--- a/src/arfs_trx_data_types.ts
+++ b/src/arfs_trx_data_types.ts
@@ -186,7 +186,7 @@ export class ArFSPrivateFileMetadataTransactionData extends ArFSFileMetadataTran
 				})
 			)
 		);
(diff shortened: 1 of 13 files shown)


src/arfsdao.ts
```diff
@@ -209,12 +217,13 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		super(arweave, appName, appVersion);
 	}
 
-	async createPublicFolder(
-		folderName: string,
-		driveId: DriveID,
-		parentFolderId?: FolderID,
+	async createPublicFolder({
+		folderData,
```
fedellen, line 221: Interesting that only create folder re-uses the data object created for estimation.

fedellen (approved):
Looking good -- sent up a boosted drive this morning 🚀

src/arfs_trx_data_types.ts
```diff
@@ -175,7 +186,7 @@ export class ArFSPrivateFileMetadataTransactionData extends ArFSFileMetadataTran
 				})
 			)
 		);
-		return new ArFSPrivateFileMetadataTransactionData(cipher, cipherIV, data);
+		return new ArFSPrivateFileMetadataTransactionData(cipher, cipherIV, fileKey, data);
```
arielmelendez, line 189: THESE PARAMS ARE OUT OF ORDER

arielmelendez: Handled in 913d2c8

vilenarios (approved):
The dry run looks straight forward enough to me.

FileKey also look to be used properly here.

Fee boosting is also seemingly correct and what I would expect.

agsuy (approved):

PR #48, opened 2021-09-27
PE-321: Folder info command

matibat:


src/arfsdao.ts
```diff
@@ -57,6 +57,7 @@ import {
 	CipherIV
 } from './types';
 import { CreateTransactionInterface } from 'arweave/node/common';
+import { cliArweave } from '.';
```
arielmelendez, line 60: We can't do this.

arielmelendez: Resolved in 327305e

arielmelendez (approved):

PR #50, opened 2021-09-28
PE-456: Assert parameters in conjunction

matibat:


arielmelendez (changes requested):
Can you add one or two contextual examples to this PR to prove out how the API would be used in practice?

matibat:
> Can you add one or two contextual examples to this PR to prove out how the API would be used in practice?

You can now check c4528e46a8948ac80a9317ec3b21f445934418f5 for an example

fedellen:
Looks pretty straight forward. Some of the parameters already have their forbidden members set up.

A question though, if we wanted certain params to have different forbidden/required combinations, would we just have to define more params? 

For a hypothetical example, if I wanted the DriveID to have forbidden members and a description unique to only the upload-file command -- would I have to define a new param like: `UploadFileDriveIdParam`  `DriveIdParam`

matibat:
> Looks pretty straight forward. Some of the parameters already have their forbidden members set up.
> 
> A question though, if we wanted certain params to have different forbidden/required combinations for specific commands, would we just have to define more params?
> 
> For a hypothetical example, if I wanted the DriveID to have forbidden members and a description unique to only the upload-file command -- would I have to define a new param like: `UploadFileDriveIdParam` along with the default implementation: `DriveIdParam`

Excellent suggestion! It was easy to add a union type to the `parameters` array to accept a custom configuration.
This example makes no sense, but you would use it like so:
```ts
new CLICommand({
	name: 'get-address',
	parameters: [
		{ name: WalletFileParameter, required: true }, // customized to be required
		SeedPhraseParameter
	],
	async action(options) {
		const context = new CommonContext(options, cliWalletDao);
```

Commit refference: 1b58709

src/CLICommand/cli_command.ts
```diff
@@ -25,8 +25,23 @@ program.addHelpCommand(false);
  */
 function setCommanderCommand(commandDescriptor: CommandDescriptor, program: CliApiObject): void {
 	let command: CliApiObject = program.command(commandDescriptor.name);
-	commandDescriptor.parameters.forEach((parameterName) => {
-		const parameter = new Parameter(parameterName);
+	commandDescriptor.parameters.forEach((p) => {
```
arielmelendez, line 28: Please use a more descriptive name than 'p'. I'm surprised the linter allowed it.

arielmelendez (changes requested):
This code is becoming increasingly complex and difficult to read and comprehend. Please add more unit tests before merging this. You. may want to reintegrate with https://github.com/ardriveapp/ardrive-cli/pull/55 first.

matibat:
@arielmelendez: I simplified the code, reduced some complexity by literally removing one of those monstrous functions, and added tests for asserting the in-conjunction parameters.
Let's see if that's better :)

src/CLICommand/parameter.ts
```diff
@@ -1,23 +1,36 @@
 export type ParameterName = string;
 export type ParameterType = 'single-value' | 'boolean' | 'array';
+export type ParameterOverridenConfig = Partial<ParameterConfig> & Pick<ParameterConfig, 'name'>;
+export type ParameterOverridingConfigWithoutName = Partial<ParameterConfig> & Omit<ParameterConfig, 'name'>;
```
matibat, line 4: No longer being used. Remove.
```suggestion
```

src/CLICommand/cli_command.test.ts
```diff
@@ -51,12 +55,49 @@ describe('CLICommand class', () => {
...
+		const commandDescriptorRequiredWallet = {
+			name: testingCommandName,
+			parameters: [
+				WalletFileParameter,
+				{ name: DrivePasswordParameter, requiredConjunctionParameters: [WalletFileParameter] }
+			],
+			async action() {
+				// eslint-disable-next-line no-console
+				console.log('DUMMY ACTION');
+			}
```
matibat, line 72: Move these to a constant

src/CLICommand/cli_command.test.ts
```diff
@@ -51,12 +55,49 @@ describe('CLICommand class', () => {
...
+			name: testingCommandName,
+			parameters: [WalletFileParameter, SeedPhraseParameter],
+			async action() {
+				// eslint-disable-next-line no-console
+				console.log('DUMMY ACTION');
+			}
+		};
+		const parsedCommandOptionsBothEspecified = {
+			[WalletFileParameter]: 'non-empty value',
+			[SeedPhraseParameter]: 'non-empty value'
```
matibat, line 94: make these to be constants

src/CLICommand/parameter.ts
```diff
@@ -40,24 +52,36 @@ export class Parameter implements ParameterData {
...
 	}
 
-	public static get(parameterName: ParameterName): ParameterData {
+	public static get(parameterName: ParameterName): ParameterConfig {
 		const param = Parameter.parameters.find((p) => p.name === parameterName);
 		if (!param) {
 			throw new Error(`No such parameter ${parameterName}`);
 		}
-		return param;
+		return Object.assign({}, param);
```
matibat, line 85:
```suggestion
		// It uses Object.assign in order to return a copy of the instance instead of the original one, so modifying it won't affect the original
		return Object.assign({}, param);
```

src/parameter_declarations.ts
```diff
@@ -143,7 +143,8 @@ Parameter.declare({
 		• parent folder ID (optional)
 			• --parent-folder-id used, otherwise
 			• all parent folder IDs should reside in the same drive
-		• Can NOT be used in conjunction with --local-file-path`
+		• Can NOT be used in conjunction with --local-file-path`,
+	forbiddenConjunctionParameters: [LocalFilePathParameter]
```
matibat, line 147: We could generate the "Can NOT be used in conjunction with ..." or "Requires to be used in conjunction with ..." messages from the `forbiddenConjunctionParameters` and `requiredConjunctionParameters`

src/CLICommand/cli_command.test.ts
matibat: a7b17024362bc0688ef26892a25f305b95ac553d

src/CLICommand/cli_command.test.ts
```diff
@@ -1,23 +1,54 @@
...
+const parsedOptionsMissingWallet = {
+	[WalletFileParameter]: undefined,
+	[DrivePasswordParameter]: nonEmptyValue
+};
+const commandDescriptorForbiddenWalletFileAndSeedPhrase: CommandDescriptor = {
+	name: testingCommandName,
+	parameters: [WalletFileParameter, SeedPhraseParameter],
+	action
+};
+const parsedCommandOptionsBothEspecified = {
```
fedellen, line 48: I think you meant `parsedCommandOptionsBothSpecified` here. Although I do like how especified sounds like especially specified

arielmelendez, line 48: Nit: `Especified` -> `Specified`

PR #51, opened 2021-09-28
PE-451: Getter function for driveKey

matibat:


arielmelendez:
Why is `.github/workflows/test_and_build.yml` deleted in this PR?

src/ardrive.ts
```diff
@@ -608,6 +612,7 @@ export class ArDrive extends ArDriveAnonymous {
 	// Provides for stubbing metadata during cost estimations since the data trx and File IDs won't yet be known
 	private async stubPrivateFileMetadata(
 		filePath: string,
+		driveKey: DriveKey,
```
arielmelendez, line 615: Having a real drive key here is not necessary.

src/ardrive.ts
```diff
@@ -441,23 +439,29 @@ export class ArDrive extends ArDriveAnonymous {
...
-	async getPrivateFolder(folderId: FolderID, drivePassword: string): Promise<ArFSPrivateFolder> {
-		const folderEntity = await this.arFsDao.getPrivateFolder(folderId, drivePassword);
+	async getPrivateFolder(folderId: FolderID, driveKey: DriveKey): Promise<ArFSPrivateFolder> {
+		const folderEntity = await this.arFsDao.getPrivateFolder(folderId, driveKey);
 		return folderEntity;
 	}
 
-	async getChildrenOfPrivateFolder(folderId: FolderID, password: string): Promise<ArFSPrivateFileOrFolderData[]> {
-		const folder = await this.arFsDao.getPrivateFolder(folderId, password);
+	async getPrivateFolderMetaData(folderId: FolderID): Promise<ArFSPrivateFolderBuilder> {
```
arielmelendez, line 452: A few problems here:
1) This function proclaims to return metadata, but in actuality it returns a builder type. Builders are NOT trustworthy objects meant for data bearing. They are intended to be short-lived objects whose brief lifespan is spent collecting fragmented data in order to build an object. So this is a confusing API that seems easy to use *incorrectly*.
2) The only usage I've found for this in the source is indeed using it incorrectly. It gets a builder (which, at construction time, doesn't do anything except store a few pieces of data and an Arweave reference) and then tries to immediately check for a driveId on it which certainly wouldn't exist yet, and which presumably would throw an exception. And in fact when I tested it, that's exactly what happened:
```
yarn ardrive list-folder  --parent-folder-id 4bf26af6-6b5d-4675-b9a7-11ca634b6d3a --wallet-file ./CLI_TEST_arweave-keyfile-g1hzNXVbh2M6LMQSUYp7HgkgxdadYqYEfw-HAajlms0.json --drive-password "mycoolpassword123"
(node:33978) UnhandledPromiseRejectionWarning: Error: Could not get driveId of private folder
```
The ArDrive layer should probably have nothing to do with ArFS-type builders. Can you please explain what you were trying to accomplish?

src/arfsdao.ts
```diff
@@ -137,27 +137,32 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 	}
 
 	async getPublicDrive(driveId: string): Promise<ArFSPublicDrive> {
+		const driveBuilder = await this.getPublicDriveMetaData(driveId);
+		return driveBuilder.build();
+	}
+
+	async getPublicDriveMetaData(driveId: string): Promise<ArFSPublicDriveBuilder> {
```
arielmelendez, line 144: Same comment here about a confusing API that's leaking builders without explanation.

src/arfsdao.ts
```diff
@@ -137,27 +137,32 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
...
-
 		const driveBuilder = new ArFSPublicDriveBuilder(driveId, this.arweave);
-		return driveBuilder.build();
+		return driveBuilder;
 	}
 
 	async getPublicFolder(folderId: string): Promise<ArFSPublicFolder> {
+		const folderBuilder = await this.getPublicFolderMetaData(folderId);
+		return await folderBuilder.build();
+	}
```
arielmelendez, line 164: Needs whitespace between functions (although I'm not convinced these functions should have been broken up in the first place).

src/utils/arfs_builders/arfs_builders.ts
```diff
@@ -25,7 +25,7 @@ export abstract class ArFSMetadataEntityBuilder<T extends ArFSEntity> {
 	constructor(protected readonly entityId: EntityID, protected readonly arweave: Arweave) {}
 
 	abstract getGqlQueryParameters(): GQLTagInterface[];
-	protected abstract buildEntity(): Promise<T>;
+	protected abstract buildEntity(driveKey?: DriveKey): Promise<T>;
```
arielmelendez, line 28: This is leaking subtype implementation details. We can't do this. The drive keys should be passed to the constructors of private subtypes only and cached for later usage.

src/arfsdao.ts
```diff
@@ -197,7 +202,7 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 				const { node } = edge;
 				cursor = edge.cursor;
 				const fileBuilder = await ArFSPublicFileBuilder.fromArweaveNode(node, this.arweave);
-				return fileBuilder.build(node);
+				return fileBuilder.build();
```
arielmelendez, line 205: Why did you get rid of the node data? Without the node data, it will cause a whole new wasted call for GQL data :(

arielmelendez (changes requested):
Things I like about this PR:
• Removing boilerplate for key generation from a lot of places. I think that's going to be super helpful to the architecture long-term.

Things that need work:
• Leaking of implementation details from lower layers
• Things that have broken functionality (and were easily manually testable given a quick risk impact analysis)
• Things that make the code less efficient (i.e. redundant network calls).

Let's try to catch up on this PR in the morning at some point.

matibat:
Some files in this PR were deleted after a bad merge, instead of rebase, I created another PR on which i cherry-picked the initial commit and manually made the merge again: see #54

PR #54, opened 2021-10-01
PE-451: getter for driveKey

matibat:


src/commands/list_folder.ts
```diff
@@ -46,7 +46,9 @@ new CLICommand({
 				CLI_APP_NAME,
 				CLI_APP_VERSION
 			);
-			children = await arDrive.listPrivateFolder(folderId, password);
+			const driveId = await arDrive.getDriveIdForFolderId(folderId);
```
matibat, line 49: This fires one extra query to get the `driveID` for the specified folder

fedellen: Is there a way for the user to provide the driveKey rather than their password?

matibat: There is. I'll mention you at that line

src/CLICommand/common_context.ts
```diff
@@ -50,6 +51,24 @@ export class CommonContext {
 		throw new Error('No wallet file neither seed phrase provided!');
 	}
 
+	public async getDriveKey(driveId: DriveID): Promise<Buffer> {
+		const driveKey = this.getParameterValue(DriveKeyParameter);
+		if (driveKey) {
+			return Buffer.from(driveKey);
+		}
```
matibat, line 58: Here @fedellen: if the user provides its drive key, the wallet and password won't be checked

src/commands/list_folder.ts
fedellen: Thats true there, but we only enter this `if (wallet && password)`. So when I provide a wallet && driveKey, we'd attempt the public version here

matibat: Got ya! I didn't realize that, here the change: 60932cf33681a2b7e1f35043ec96c7d6462908fd

arielmelendez (approved):
Love the cleanup!

PR #55, opened 2021-10-01
PE-451: Implement Context.getRequiredParameterValue()

matibat:


src/commands/create_drive.ts
```diff
@@ -25,15 +24,15 @@ new CLICommand({
 		DryRunParameter
 	],
 	async action(options) {
-		const context = new CommonContext(options, cliWalletDao);
-		const wallet: Wallet = await context.getWallet();
+		const parameters = new ParametersHelper(options, cliWalletDao);
```
fedellen, line 27: Looks like we don't need to supply a walletDao here anymore now that you're using it as a default

fedellen: There are a couple similar cases below

src/index.ts
```diff
@@ -31,32 +31,39 @@ export const cliArweave = Arweave.init({
...
-		new ArDriveCommunityOracle(arweave),
-		CLI_APP_NAME,
-		CLI_APP_VERSION,
-		priceEstimator,
-		feeMultiple,
-		dryRun
-	);
+export function arDriveFactory(settings?: ArDriveSettingsAnonymus): ArDriveAnonymous;
+export function arDriveFactory(settings: ArDriveSettings): ArDrive;
+export function arDriveFactory(s?: ArDriveSettingsAnonymus): ArDrive | ArDriveAnonymous {
```
fedellen, line 47: Nice overload example!

src/CLICommand/parameters_helper.ts
```diff
@@ -51,6 +53,10 @@ export class CommonContext {
 		throw new Error('No wallet file neither seed phrase provided!');
 	}
 
+	public async getWalletAddress(): Promise<string> {
+		return this.getParameterValue(DriveAddressParameter) || this.getWallet().then((wallet) => wallet.getAddress());
```
arielmelendez, line 57: `--drive-address` was never outlined in the man page and honestly I find its name confusing. Is it a drive entity ID? A drive tx ID? It's actually neither. It's a wallet address passed in by a user or that of a supplied wallet file. I'm going to rename this accordingly (i.e. to the `--address` option that was specified in the man page).

arielmelendez: I also just noticed a misspelling in the `DestinationAddressParameter` value assignment:
```
export const DestinationAddressParameter = 'destAddres'; // This is missing a trailing 's'
```
and that seems to be used in `send-ar` action. I'll fix that as well.

arielmelendez: af53e1c and 771aae0

src/commands/create_drive.ts
arielmelendez: 34aa156

src/CLICommand/parameters_helper.ts
arielmelendez: Done

src/commands/create_drive.ts
arielmelendez: Took care of it in 34aa156

fedellen (approved):
Looks good!

PR #65, opened 2021-10-07
PE-518: Max depth

matibat:


src/arfsdao.ts
```diff
@@ -276,7 +276,7 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 		maxDepth: number,
 		includeRoot: boolean
 	): Promise<ArFSPublicFileOrFolderWithPaths[]> {
-		if (maxDepth < 0 || !Number.isInteger(maxDepth)) {
+		if (maxDepth < 0 || (!Number.isInteger(maxDepth) && maxDepth !== Number.POSITIVE_INFINITY)) {
```
matibat, line 279: I used `Number.POSITIVE_INFINITY` as `maxDepth` when `--all` entities are requested. Because of that, I had to add this check to the condition. But now that we do have `ParametersHelper.getMaxDepth(): number` we could assert this in the input-reading step.

Do you guys think it's better to move this check to the `ParametersHelper.getMaxDepth()` so it can be removed here?

src/CLICommand/parameters_helper.ts
```diff
@@ -82,6 +84,15 @@ export class ParametersHelper {
 		throw new Error(`No drive key or password provided!`);
 	}
 
+	public async getMaxDepth(): Promise<number> {
+		if (await this.getParameterValue(AllParameter)) {
+			return Number.POSITIVE_INFINITY;
+		} else {
+			const value = Number(await this.getParameterValue(MaxDepthParameter));
+			return value ? value : 0;
```
matibat, line 92:
```suggestion
			if(!Number.isInteger(value) || value < 0) {
				throw new Error('maxDepth should be a non-negative integer!')
			}
			return value;
```

src/arfsdao.ts
```diff
@@ -276,7 +276,7 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
...
 			throw new Error('maxDepth should be a non-negative integer!');
 		}
```
matibat, line 281:
```suggestion
```

src/arfsdao.ts
```diff
@@ -850,7 +850,7 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		maxDepth: number,
 		includeRoot: boolean
 	): Promise<ArFSPrivateFileOrFolderWithPaths[]> {
-		if (maxDepth < 0 || !Number.isInteger(maxDepth)) {
+		if (maxDepth < 0 || (!Number.isInteger(maxDepth) && maxDepth !== Number.POSITIVE_INFINITY)) {
 			throw new Error('maxDepth should be a non-negative integer!');
 		}
```
matibat, line 855:
```suggestion
```

src/arfsdao.ts
matibat: I already changed this :)

src/CLICommand/parameters_helper.test.ts
```diff
@@ -364,4 +369,12 @@ describe('ParametersHelper class', () => {
 			CLICommand.parse(program, [...baseArgv, testCommandName]);
 		});
 	});
+
+	describe('getMaxDepth method', () => {
+		it(`Defaults to zero`);
+		it(`Does not accept a decimal`);
+		it(`Does not accept a negative integer`);
+		it(`Max depth is infinity when --all is specified`);
```
arielmelendez, line 377: Don't use infinity. In fact we should error if the user provides infinity and test for that.

src/CLICommand/parameters_helper.ts
```diff
@@ -82,6 +84,18 @@ export class ParametersHelper {
 		throw new Error(`No drive key or password provided!`);
 	}
 
+	public async getMaxDepth(): Promise<number> {
+		if (await this.getParameterValue(AllParameter)) {
+			return Number.POSITIVE_INFINITY;
```
arielmelendez, line 89: Use Number.MAX_SAFE_INTEGER

src/CLICommand/parameters_helper.test.ts
arielmelendez: Actually it looks like Number.isInteger fails for POSITIVE_INFINITY

src/arfsdao.ts
```diff
@@ -276,9 +276,6 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 		maxDepth: number,
 		includeRoot: boolean
 	): Promise<ArFSPublicFileOrFolderWithPaths[]> {
-		if (maxDepth < 0 || !Number.isInteger(maxDepth)) {
```
arielmelendez, line 279: Why remove these guards? If you had left them in, you'd have seen that using POSITIVE_INFINITY doesn't actually work 🤔

src/arfsdao.ts
```diff
@@ -850,9 +847,6 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		maxDepth: number,
 		includeRoot: boolean
 	): Promise<ArFSPrivateFileOrFolderWithPaths[]> {
-		if (maxDepth < 0 || !Number.isInteger(maxDepth)) {
```
arielmelendez, line 853: Why remove these guards? If you had left them in, you'd have seen that using POSITIVE_INFINITY doesn't actually work 🤔

src/parameter_declarations.ts
```diff
@@ -192,7 +192,8 @@ Parameter.declare({
 	name: AllParameter,
 	aliases: ['--all'],
 	description: `(OPTIONAL) gets all contents within this folder, including child files/folders`,
-	type: 'boolean'
+	type: 'boolean',
+	forbiddenConjunctionParameters: [MaxDepthParameter]
```
arielmelendez, line 196: Does it make sense to also list the --all parameter as forbidden in conjunction with the all parameter? Is it functionally different to mark them forbidden in one direction or the other?

src/arfsdao.ts
arielmelendez: While that benefits the CLI layer, it removes the safety checks here for ArDrive Core itself (once ArFSDao ultimately lives there).

PR #66, opened 2021-10-07
feat(move folder): Implement move folder PE-320

fedellen:
This PR implements the move folder command.

Notably, I have temporarily renamed the `ParentFolderId` shortcut parameter to `-pf` to avoid conflicts with the folderId param. This change also allows the file-id param shortcut to use `-f` once again.

### DIFF
--- a/src/arfsdao.ts
+++ b/src/arfsdao.ts
@@ -98,12 +98,12 @@ export interface ArFSMovePrivateFileResult extends ArFSMoveFileResult {
 	fileKey: FileKey;
 }
 
-export interface ArFSMoveFolderResult {
+export interface ArFSMovePublicFolderResult {
 	metaDataTrxId: TransactionID;
 	metaDataTrxReward: TransactionID;
 }
 
-export interface ArFSMovePrivateFolderResult extends ArFSMoveFolderResult {
+export interface ArFSMovePrivateFolderResult extends ArFSMovePublicFolderResult {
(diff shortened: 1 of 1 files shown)


src/ardrive.ts
```diff
@@ -217,29 +217,29 @@ export class ArDrive extends ArDriveAnonymous {
 	async movePublicFile(fileId: FileID, newParentFolderId: FolderID): Promise<ArFSResult> {
 		const driveId = await this.getDriveIdAndAssertDrive(newParentFolderId);
 
-		const originalFileMetaData = await this.getPublicFile(fileId);
```
arielmelendez, line 220: Were all these naming changes in the move file functions necessary in this PR? It seems like you may have been going down a path to unify file and/or folder moving and ended up going with separate functions...

src/arfsdao.ts
```diff
@@ -98,6 +98,15 @@ export interface ArFSMovePrivateFileResult extends ArFSMoveFileResult {
 	fileKey: FileKey;
 }
 
+export interface ArFSMoveFolderResult {
```
arielmelendez, line 101: It could make sense to make a useful distinction between public and private here in order to reduce the chance of misuse in the future. Consider `ArFSMovePublicFolderResult`.

arielmelendez (approved):
Approved with nits on superficial things. Just double check on that -pf flag.

PR #70, opened 2021-10-09
PE-515: Integration Test the ArDrive Class

arielmelendez:
Uses stubs for all outbound network activity (i.e. getter functions) but tests most classes in integration during ArDrive writing workflows.

Getter workflows not tested due to lack of a mock server for Arweave, smartweave oracles, etc.

### DIFF
--- a/tests/integration/ardrive.int.test.ts
+++ b/tests/integration/ardrive.int.test.ts
@@ -157,9 +157,7 @@ describe('ArDrive class', () => {
 	describe('utility function', () => {
 		describe('sendCommunityTip', () => {
 			it('returns the correct TipResult', async () => {
-				stub(communityOracle, 'selectTokenHolder').callsFake(() => {
-					return Promise.resolve(stubArweaveAddress);
-				});
+				stub(communityOracle, 'selectTokenHolder').resolves(stubArweaveAddress);
 
 				const result = await arDrive.sendCommunityTip('12345');
 
@@ -173,12 +171,12 @@ describe('ArDrive class', () => {
 	});
 
 	describe('drive function', () => {
+		beforeEach(() => {
+			stub(walletDao, 'walletHasBalance').resolves(true);
+		});
+
 		describe('createPublicDrive', () => {
 			it('returns the correct ArFSResult', async () => {
-				stub(walletDao, 'walletHasBalance').callsFake(() => {
-					return Promise.resolve(true);
-				});
-
 				const result = await arDrive.createPublicDrive('TEST_DRIVE');
 				assertCreateDriveExpectations(result, 75, 21);
 			});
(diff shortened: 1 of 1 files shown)


src/utils/ardrive.test.ts
```diff
@@ -0,0 +1,235 @@
...
+			inputsAndExpectedOutputs.forEach(([input, expectedOutput]) => {
+				expect(arDrive.encryptedDataSize(input)).to.equal(expectedOutput);
+			});
+		});
+	});
+
+	describe('getTipTags function', () => {
+		it('returns the expected tags', () => {
+			const baseTags = [
+				{ name: 'App-Name', value: 'ArDrive-CLI' },
```
arielmelendez, line 97: TODO: Should this be ArDrive-Core in the absence of passed in values?

arielmelendez:
@fedellen:
> Should we also move the file/folder if the new parent folder id is the same as its current parent folder? (i.e: send up identical metadata with a new unix time)

We need to check that the recursive functions that follow the folder tree structure in FolderTreeHierarchy won't crap out in the case you mentioned and any time there's a circular reference in the tree, i.e. `GrandChild -> Parent -> Grandparent -> Grandchild`

src/ardrive.ts
```diff
@@ -573,7 +593,14 @@ export class ArDrive extends ArDriveAnonymous {
...
 	encryptedDataSize(dataSize: ByteCount): ByteCount {
-		return (dataSize / 16 + 1) * 16;
+		if (dataSize < 0 || !Number.isInteger(dataSize)) {
+			throw new Error(`dataSize must be non-negative, integer value! ${dataSize} is invalid!`);
+		}
+		if (dataSize > Number.MAX_SAFE_INTEGER - 16) {
+			throw new Error(`Max unencrypted dataSize allowed is ${Number.MAX_SAFE_INTEGER - 16}!`);
+		}
+		const modulo16 = dataSize % 16;
+		return dataSize - modulo16 + 16;
```
fedellen, line 603: Whats the difference here? And I wonder why it worked with your `file-size` test command code before.

tests/integration/ardrive.int.test.ts
```diff
@@ -0,0 +1,529 @@
...
+) {
+	// Ensure that 1 arfs entity was created
+	expect(result.created.length).to.equal(1);
+
+	// Ensure that the file data entity looks healthy
+	const fileEntity = result.created[0];
+	expect(fileEntity.dataTxId).to.match(trxIdRegex);
+	expect(fileEntity.entityId).to.match(entityIdRegex);
+	switch (drivePrivacy) {
+		case 'public':
```
fedellen, line 469: A switch for basically a boolean value might be a bit much

tests/integration/ardrive.int.test.ts
```diff
@@ -0,0 +1,529 @@
...
+				});
+				stub(arfsDao, 'getDriveIdForFolderId').callsFake(() => {
+					return Promise.resolve(stubEntityID);
+				});
+				stub(arfsDao, 'getPublicDrive').callsFake(() => {
+					return Promise.resolve(stubPublicDrive);
+				});
+				stub(arfsDao, 'getPublicFolder').callsFake(() => {
+					return Promise.resolve(stubPublicFolder);
+				});
```
fedellen, line 308: We do need to stub the community oracle functions when we upload files. I think we're probably hitting the network here

PR #81, opened 2021-10-12
PE-542 Consolidate upload file functions

arielmelendez:


src/arfsdao.ts
```diff
@@ -487,65 +524,33 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
-			dataContentType,
-			fileId,
-			driveKey
-		);
-		const fileMetadataPrototype = new ArFSPrivateFileMetaDataPrototype(
-			fileMetaData,
-			driveId,
-			fileId,
-			parentFolderId
+		return this.uploadFile(
```
fedellen, line 527: Cool, this is inferring the generic types to pass through here. I noticed there there was no types here `uploadFile<myTypes>`

PR #83, opened 2021-10-12
Gracefully catch in-conjunction errors

matibat:


src/CLICommand/cli_command.ts
```diff
@@ -47,8 +47,10 @@ function setCommanderCommand(commandDescriptor: CommandDescriptor, program: CliA
 		}
 	});
 	command = command.action(async (options) => {
-		assertConjunctionParameters(commandDescriptor, options);
-		await commandDescriptor.action(options).catch((err) => {
+		await (async function () {
```
fedellen, line 50: So this will now print the error from `assertConjunctionParameters` rather than a call stack? Or whats before and after behavior change here?

matibat: It will only print the error message, yes. Previously it thrown an uncaught exception.

I think we could make it to print the whole stack if there's the `--verbose` flag set :thinking:

fedellen (approved):

src/CLICommand/cli_command.ts
fedellen: This fixes an issue Agustin is running into. Looks good to me

Sorry for the confusing question. I was just wondering if I was missing something here

matibat:
https://ardrive.atlassian.net/browse/PE-549

PR #85, opened 2021-10-12
PE-545: Additional error cases on move-entity

fedellen:
This PR adds new error cases for the move entity functions. Specifically, users can no longer move ~~their root folder~~ a parent folder into any of its children folders and no longer move a file/folder if the supplied newParentFolderId matches its current parent folder ID


src/arfsdao.ts
```diff
@@ -709,6 +709,21 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		return latestRevisionsOnly ? allFiles.filter(latestRevisionFilter) : allFiles;
 	}
 
+	async getChildrenFolderIds(
+		folderId: FolderID,
+		allFolderEntitiesOfDrive: ArFSPublicFolder[] | ArFSPrivateFolder[]
+	): Promise<FolderID[]> {
+		const hierarchy = FolderHierarchy.newFromEntities(allFolderEntitiesOfDrive);
+		return hierarchy.folderIdSubtreeFromFolderId(folderId, Number.POSITIVE_INFINITY);
```
matibat, line 717: `maxDepth` should'n be greater than `Number.MAX_SAFE_INTEGER`
```suggestion
		return hierarchy.folderIdSubtreeFromFolderId(folderId, Number.MAX_SAFE_INTEGER);
```

src/arfsdao.ts
```diff
@@ -709,6 +709,21 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		return latestRevisionsOnly ? allFiles.filter(latestRevisionFilter) : allFiles;
 	}
 
+	async getChildrenFolderIds(
+		folderId: FolderID,
+		allFolderEntitiesOfDrive: ArFSPublicFolder[] | ArFSPrivateFolder[]
```
arielmelendez, line 714: You can use the parent type here to consolidate types

src/arfsdao.ts
```diff
@@ -709,6 +709,21 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+	): Promise<FolderID[]> {
+		const hierarchy = FolderHierarchy.newFromEntities(allFolderEntitiesOfDrive);
+		return hierarchy.folderIdSubtreeFromFolderId(folderId, Number.MAX_SAFE_INTEGER);
+	}
+
+	async getPrivateChildrenFolderIds(folderId: FolderID, driveId: DriveID, driveKey: DriveKey): Promise<FolderID[]> {
```
arielmelendez, line 720: Consider deconstructing params from a type to avoid using FolderID and DriveID in the wrong positions.

src/arfsdao.ts
matibat: The algorithm right now is safe because:
```ts
maxDepth = Number.POSITIVE_INFINITY;
maxDepth = maxDepth - 1; // Still being POSITIVE_INFINITY
maxDepth > 0; // true
```

But it could be broken in the future, so this is changed at PR #90.

tests/integration/ardrive.int.test.ts
```diff
@@ -225,29 +133,75 @@ describe('ArDrive class - integrated', () => {
...
 				stub(arfsDao, 'getPublicDrive').resolves(stubPublicDrive);
 				stub(arfsDao, 'getPublicFolder').resolves(stubPublicFolder);
+				stub(arfsDao, 'getPublicChildrenFolderIds').resolves([stubEntityIDAltTwo]); // Uses Alt ID Two
+			});
+
+			it('throws an error if its being moved inside any of its children folders', async () => {
+				await expectAsyncErrorThrow({
+					promiseToError: arDrive.movePublicFolder({
+						folderId: stubEntityID,
+						newParentFolderId: stubEntityIDAltTwo // Alt ID Two will be in Children folder ids
```
arielmelendez, line 146: I think we should create a set of stubs that are known to have a hierarchical relationship and access them in a clear way, e.g.:
```
interface StubFolderHierarchy<T extends ArFSFileOrFolderEntity> {
    root: T;
    child: T;
    grandchild: T;
}
const stubFolderHierarchy = {
   root: ...,
    child: ...,
    grandchild: ...
}
```
and use like:
```
stubFolderHierarchy.root // do whatever
```

arielmelendez (approved):

tests/integration/ardrive.int.test.ts
```diff
@@ -132,150 +135,281 @@ describe('ArDrive class - integrated', () => {
...
+					folderHierarchy.parentFolder.entityId,
+					folderHierarchy.childFolder.entityId,
+					folderHierarchy.grandChildFolder.entityId
+				]);
+
 				await expectAsyncErrorThrow({
 					promiseToError: arDrive.movePublicFolder({
-						folderId: stubEntityID,
-						newParentFolderId: stubEntityIDAltTwo // Alt ID Two will be in Children folder ids
+						folderId: folderHierarchy.rootFolder.entityId,
```
arielmelendez, line 155: I would have expected this to throw for being a root folder that got moved, NOT because of the child relationship.

tests/integration/ardrive.int.test.ts
```diff
@@ -132,150 +135,281 @@ describe('ArDrive class - integrated', () => {
...
+					folderHierarchy.grandChildFolder.entityId
+				]);
+
+				stub(arDrive, 'getDriveIdAndAssertDrive').resolves(stubEntityIDAlt); // Returns different Drive ID
+
 				await expectAsyncErrorThrow({
 					promiseToError: arDrive.movePublicFolder({
-						folderId: stubEntityID,
-						newParentFolderId: stubEntityID // Stub ID will MATCH current parentFolderId field
+						folderId: folderHierarchy.childFolder.entityId,
```
arielmelendez, line 204: These id choices are ok, but it confuses the test a bit because the focus should be on the driveId aspect. Here my eyes are drawn to the fact that you're moving the child folder to the parent folder that it's already in.

fedellen: d1d65e0

tests/integration/ardrive.int.test.ts
fedellen: 07acb87

src/arfsdao.ts
```diff
@@ -709,6 +718,25 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+	}
+
+	async getPrivateChildrenFolderIds({
+		folderId,
+		driveId,
+		driveKey
+	}: getPrivateChildrenFolderIdsParams): Promise<FolderID[]> {
+		return this.getChildrenFolderIds(folderId, await this.getAllFoldersOfPrivateDrive(driveId, driveKey, true));
+	}
+	async getPublicChildrenFolderIds({ folderId, driveId }: getPublicChildrenFolderIdsParams): Promise<FolderID[]> {
```
arielmelendez, line 736: nit: whitespace

src/example.test.ts
```diff
@@ -34,8 +34,8 @@ function basicInputOutputExample(input: number) {
 
 // Async input/output example function
 async function asyncInputOutputExample(input: number) {
-	// Waits 100ms to simulate async call
-	await sleep(100);
+	// Waits 1ms to simulate async call
+	await sleep(1);
```
arielmelendez, line 38: Nice

PR #86, opened 2021-10-13
PE-550: Quote the balance in AR instead of Winston

matibat:


src/commands/get_balance.ts
```diff
@@ -2,14 +2,19 @@ import { cliWalletDao } from '..';
 import { CLICommand, ParametersHelper } from '../CLICommand';
 import { AddressParameter, SeedPhraseParameter, WalletFileParameter } from '../parameter_declarations';
 
+const anArInWinston = 1_000_000_000_000;
+const numberOfDigitsAfterComma = 8;
```
arielmelendez, line 6: Why limit this to 8 digits? There's 10^12 Winston per AR. Therefore we need 12 decimal places for precision.

src/commands/get_balance.ts
```diff
@@ -2,14 +2,19 @@ import { cliWalletDao } from '..';
...
 new CLICommand({
 	name: 'get-balance',
 	parameters: [WalletFileParameter, SeedPhraseParameter, AddressParameter],
 	async action(options) {
 		const parameters = new ParametersHelper(options);
 		const address = await parameters.getWalletAddress();
-		const balance = await cliWalletDao.getAddressWinstonBalance(address);
-		console.log(balance);
+		const balanceInWinston = await cliWalletDao.getAddressWinstonBalance(address);
+		const balanceInAR = balanceInWinston / anArInWinston;
```
arielmelendez, line 15: Consider using `arweave.ar.winstonToAr` instead.

arielmelendez (changes requested):
Number of decimal places is the critical change needed.

src/commands/get_balance.ts
matibat: Here it is: 84da1c34155b75342e49264d08cf55eaf9601b8b

PR #88, opened 2021-10-13
PE-545 secure passwords

arielmelendez:
This PR will start in draft state. I've tested this so far like so:
```
yarn ardrive drive-info  --drive-id ff40ec22-77d9-4e52-baf1-ee6ef538202a -P --wallet-file ./CLI_TEST_arweave-keyfile-g1hzNXVbh2M6LMQSUYp7HgkgxdadYqYEfw-HAajlms0.json <<< <the password>

ARDRIVE_DRIVE_PW=<the password> yarn ardrive drive-info  --drive-id ff40ec22-77d9-4e52-baf1-ee6ef538202a -P --wallet-file ./CLI_TEST_arweave-keyfile-g1hzNXVbh2M6LMQSUYp7HgkgxdadYqYEfw-HAajlms0.json

yarn ardrive drive-info  --drive-id ff40ec22-77d9-4e52-baf1-ee6ef538202a -P --wallet-file ./CLI_TEST_arweave-keyfile-g1hzNXVbh2M6LMQSUYp7HgkgxdadYqYEfw-HAajlms0.json
✔ Enter drive password: … *****************
```

Looks like I've got a broken test to fix too. Just soliciting early feedback.

### DIFF
--- a/src/CLICommand/parameters_helper.ts
+++ b/src/CLICommand/parameters_helper.ts
@@ -25,13 +25,13 @@ interface GetDriveKeyParams {
 	useCache?: boolean;
 }
 
-const staticDriveKeyCache: { [key: string]: Buffer } = {};
-
 /**
  * @type {ParametersHelper}
  * A class that assists with handling Commander options during common ArDrive CLI workflows
  */
 export class ParametersHelper {
+	private static readonly driveKeyCache: { [key: string]: DriveKey } = {};
+
 	/**
 	 * @returns {ParametersHelper}
 	 * @param {any} options The object containing the parameterName: value mapping
@@ -86,7 +86,7 @@ export class ParametersHelper {
 		// • (--wallet-file or --seed-phrase) + (--unsafe-drive-password or --private password)
 
 		if (useCache) {
-			const cachedDriveKey = staticDriveKeyCache[driveId];
+			const cachedDriveKey = ParametersHelper.driveKeyCache[driveId];
 			if (cachedDriveKey) {
 				return cachedDriveKey;
 			}
@@ -95,7 +95,7 @@ export class ParametersHelper {
 		const driveKey = this.getParameterValue(DriveKeyParameter);
 		if (driveKey) {
 			const paramDriveKey = Buffer.from(driveKey, 'base64');
-			staticDriveKeyCache[driveId] = paramDriveKey;
+			ParametersHelper.driveKeyCache[driveId] = paramDriveKey;
 			return paramDriveKey;
 		}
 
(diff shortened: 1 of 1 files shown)


arielmelendez:
Interestingly I'm seeing a failing test locally but not in CI... will investigate in the morning.

src/CLICommand/parameters_helper.ts
```diff
@@ -87,6 +94,59 @@ export class ParametersHelper {
...
+				'Password not detected for private drive operation! Please provide a password via the --private option (recommended) or the --unsafe-drive-password option (not recommended).'
+			);
+			process.exit(1);
+		}
+		return unsafePassword;
+	}
+
 	public async getMaxDepth(defaultDepth: number): Promise<number> {
 		if (this.getParameterValue(AllParameter)) {
 			return Number.POSITIVE_INFINITY;
```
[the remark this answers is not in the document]
matibat: Merged from `cli_v2`: 4f01911

matibat:
Added a new method
```ts
function exitProgram(exitCode: number): void {
	process.exitCode = exitCode;
}
```
Which instead of interrupting the program execution, simply sets the `exitCode` to the one the `action()` returns (now it does, and defaults to 0 = SUCCESS)

src/CLICommand/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const SUCCES_EXIT_CODE = 0;
```
fedellen, line 1: spelling nit: SUCCES -> SUCCESS

src/commands/upload_file.ts
```diff
@@ -67,13 +59,10 @@ new CLICommand({
 			const singleParameter = {
 				parentFolderId: options.parentFolderId,
 				wrappedEntity: wrapFileOrFolder(options.localFilePath),
-				destinationFileName: options.destFileName,
-				drivePassword: options.drivePassword,
-				driveKey: options.driveKey
```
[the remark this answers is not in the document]
arielmelendez: Handled in efc2d99 , though I worry about our user story for mixing public and private drives in the same CSV. Perhaps we ignore the -P flag and rely instead on CSV parameters? i.e. on line 97
```
						if (await parameters.getIsPrivate()) {
```
perhaps we just check if a driveKey or drivePassword are provided on the file params.
If we did this, we might want to change the upload-file command to something like bulk-upload-file to make sure the parameters are harmonious. Any other ideas?

src/CLICommand/parameters_helper.ts
```diff
@@ -19,6 +19,14 @@ import passwordPrompt from 'prompts';
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type ParameterOptions = any;
 
+interface GetDriveKeyParams {
+	driveId: DriveID;
+	drivePassword?: string;
+	useCache?: boolean;
+}
+
+const staticDriveKeyCache: { [key: string]: Buffer } = {};
```
matibat, line 28: I would make it a `private static` attribute of ParametersHelper rather than a global-scoped constant

matibat: Also, TypeScript does allow us to use correct here, it's only with the key that it gets mad
```suggestion
const staticDriveKeyCache: { [key: string]: DriveKey } = {};
```

fedellen (approved):
Looks good!

PR #91, opened 2021-10-13
feat(list drives): Initialize list drives command PE-547

fedellen:
This PR addes the `list-drives` command which will gracefully fetch all of the drive entities for a supplied wallet or arweave address. 

This command operates using only the anonymous ArDrive class. Private drives will only have their `name` and `rootFolderId`  fields decrypted if the provided drive key or drive password succeeds in decrypting the dataJSON. Otherwise they will output:

```shell
"name": "ENCRYPTED",
"rootFolderId": "ENCRYPTED"
```

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -1,6 +1,6 @@
 import { ArFSDAO, PrivateDriveKeyData } from './arfsdao';
 import { CommunityOracle } from './community/community_oracle';
-import { deriveDriveKey, DrivePrivacy, GQLTagInterface, winstonToAr } from 'ardrive-core-js';
+import { ArFSDriveEntity, deriveDriveKey, DrivePrivacy, GQLTagInterface, winstonToAr } from 'ardrive-core-js';
 import {
 	TransactionID,
 	ArweaveAddress,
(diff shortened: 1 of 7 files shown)


src/parameter_declarations.ts
```diff
@@ -83,8 +83,8 @@ Parameter.declare({
 Parameter.declare({
 	name: AddressParameter,
 	aliases: ['-a', '--address'],
-	description: 'the address',
-	forbiddenConjunctionParameters: [DrivePasswordParameter, DriveKeyParameter]
+	description: 'the address'
+	// forbiddenConjunctionParameters: [DrivePasswordParameter, DriveKeyParameter]
```
matibat, line 87: It didn't make sense to have this restriction in the first place, right?
```suggestion
```

src/arfsdao_anonymous.ts
```diff
@@ -74,6 +75,61 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
...
+		if (!driveId) {
+			throw new Error('Drive-Id tag missing or corrupted!!');
+		}
+		return driveId;
+	}
+
+	async getAllDrivesForAddress(
+		address: ArweaveAddress,
+		privateKeyData: PrivateKeyData,
+		latestRevisionsOnly = false
```
arielmelendez, line 94: Should be true by default?

src/ardrive.ts
```diff
@@ -147,6 +148,14 @@ export class ArDriveAnonymous extends ArDriveType {
 		return this.arFsDao.getPublicFile(fileId);
 	}
 
+	async getAllDrivesForAddress(address: ArweaveAddress, privateKeyData: PrivateKeyData): Promise<ArFSDriveEntity[]> {
+		return this.arFsDao.getAllDrivesForAddress(address, privateKeyData);
+	}
+
+	async getLatestDriveIdForAddress(address: ArweaveAddress): Promise<DriveID> {
```
arielmelendez, line 155: Excise

PR #97, opened 2021-10-18
PE-586: Enforce unique folder name

fedellen:
This PR adds a check for conflicting names within the supplied parent folder ID on the `move-folder` and `create-folder` commands

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -1301,4 +1301,8 @@ export class ArDrive extends ArDriveAnonymous {
 			)
 		);
 	}
+
+	async assertValidPassword(password: string): Promise<void> {
+		await this.arFsDao.assertValidPassword(password);
+	}
 }
--- a/src/arfsdao.ts
+++ b/src/arfsdao.ts
@@ -34,7 +34,7 @@ import {
 	ArFSPublicFileMetadataTransactionData,
 	ArFSPublicFolderTransactionData
 } from './arfs_trx_data_types';
-import { buildQuery } from './query';
+import { ASCENDING_ORDER, buildQuery } from './query';
 import { ArFSFileToUpload } from './arfs_file_wrapper';
 import {
 	DriveID,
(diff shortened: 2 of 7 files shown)


src/ardrive.ts
```diff
@@ -341,6 +341,12 @@ export class ArDrive extends ArDriveAnonymous {
 			throw new Error(errorMessage.cannotMoveIntoSamePlace('Folder', newParentFolderId));
 		}
 
+		// Assert that there are no duplicate names in the destination folder
+		const entityNamesInParentFolder = await this.arFsDao.getPublicChildNamesOfParentFolderId(newParentFolderId);
```
arielmelendez, line 345: How about a simpler name? Maybe:
`getNamesOfPublicEntitiesWithParentFolderId`
or
`getPublicEntityNamesInFolder`

arielmelendez: Same comment applies to the private use case.

src/error_message.ts
```diff
@@ -1,6 +1,7 @@
 import { FolderID } from './types';
 
 export const errorMessage = {
+	cannotUseDuplicateNameInParentFolder: 'Parent folder already has an entity with the name!',
```
arielmelendez, line 4: How about a constant name like `entityNameExists` and a message like `Entity name already exists in destination folder!"

tests/integration/ardrive.int.test.ts
```diff
@@ -112,21 +112,57 @@ describe('ArDrive class - integrated', () => {
 
 	describe('folder function', () => {
 		describe('createPublicFolder', () => {
+			beforeEach(() => {
+				stub(arfsDao, 'getPublicChildNamesOfParentFolderId').resolves(['CONFLICTING_NAME']);
+			});
+
+			it('throws an error if the folder name conflicts with another folder in the parent', async () => {
```
arielmelendez, line 119: I find `in the parent` to be ambiguous. "The parent" could refer to two different things: the origin parent or destination parent. I think it's easy enough to say "the destination folder".

src/arfsdao.ts
```diff
@@ -718,6 +726,85 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+			node: GQLNodeInterface,
+			entityType: 'file' | 'folder'
+		) => ArFSFileOrFolderBuilder<ArFSFileOrFolderEntity>,
+		latestRevisionsOnly = true
+	): Promise<ArFSFileOrFolderEntity[]> {
+		let cursor = '';
+		let hasNextPage = true;
+		const allEntities: ArFSFileOrFolderEntity[] = [];
+
+		// TODO: Always derive owner of a wallet from the earliest drive a wallet by default
```
arielmelendez, line 741: Nit: Grammar
Also, we're going to make this the default behavior but at a later point will add the ability to NOT filter that way in order to find problematic (e.g. fraudulent or corrupted) entities.

PR #98, opened 2021-10-19
PE-583: same password for all drives

matibat:


fedellen (approved):

src/arfsdao.ts
```diff
@@ -793,15 +793,14 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+			tags: [
 				{ name: 'Entity-Type', value: 'drive' },
 				{ name: 'Drive-Privacy', value: 'private' }
 			],
-			undefined,
-			walletAddress,
-			DESCENDENT_ORDER
-		);
+			owner: walletAddress,
+			sort: ASCENDING_ORDER
```
matibat, line 806: Isn't the default ASCENDING? If not, we DO have to change it at `buildQuery()` args

src/arfsdao.ts
```diff
@@ -781,4 +793,34 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+			this.arweave,
+			new PrivateKeyData({ password, wallet: this.wallet as JWKWallet })
+		);
+		const safelyBuiltDrive = await safeDriveBuilder.build();
+		if (
+			safelyBuiltDrive.name === ENCRYPTED_DATA_PLACEHOLDER ||
+			safelyBuiltDrive.rootFolderId === ENCRYPTED_DATA_PLACEHOLDER
+		) {
+			throw new Error(`Invalid password! Please type the same as your other private drives!`);
+		}
```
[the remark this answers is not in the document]
fedellen: Drive key is derived within the context of PrivateKeyData with the provided pw and wallet

src/arfsdao.ts
```diff
@@ -781,4 +793,34 @@ export class ArFSDAO extends ArFSDAOAnonymous {
...
+				{ name: 'Entity-Type', value: 'drive' },
+				{ name: 'Drive-Privacy', value: 'private' }
+			],
+			owner: walletAddress,
+			sort: ASCENDING_ORDER
+		});
+		const response = await this.arweave.api.post(graphQLURL, query);
+		const { data } = response.data;
+		const { transactions } = data;
+		const { edges } = transactions;
```
matibat, line 811: It should early return if no drive is found. This means the user will be able to enter whichever the password it wants as it is the very first drive
```suggestion
		const { edges } = transactions;
		if (!edges.length) {
				return;
		}
```

PR #99, opened 2021-10-19
PE-593 get drive key cmd

arielmelendez:


src/commands/get_drive_key.ts
```diff
@@ -0,0 +1,14 @@
+import { CLICommand, ParametersHelper } from '../CLICommand';
+import { DriveCreationPrivacyParameters, DriveIdParameter } from '../parameter_declarations';
+import { urlEncodeHashKey } from '../utils';
+
+new CLICommand({
+	name: 'get-drive-key',
```
matibat, line 6: As we won't check if it is actually a valid drive key (we don't assert we are **getting** an actual drive key), we might want to rename this command to something more generic
```suggestion
	name: 'generate-drive-key',
```

matibat (approved):
Left one single suggestion. Apart from that, it looks good

PR #100, opened 2021-10-19
PE-558: Enforce unique file name

fedellen:
This PR enforces unique file names within the supplied destination folder for `upload-file` and `move-file`.

#### Expected behavior for move file:
- Errors out if destination folder has conflicting FILE or FOLDER name

#### Expected behavior for upload file performed on a single file:
- Errors out if destination folder has conflicting FOLDER name
- Uploads as a REVISION if destination folder has conflicting FILE name

#### Expected behavior for upload file performed on a folder (uploads -- for all entities within the tree):
- Folder names that conflict with a destination FILE name will error
- Folder names that conflict with destination FOLDER name will use the existing folder ID rather than creating a new folder
- File names that conflict with a destination FOLDER name will error
- File names that conflict with destination FILE name will be uploaded as a REVISION

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -587,7 +587,7 @@ export class ArDrive extends ArDriveAnonymous {
 		parentFolderId,
 		wrappedFolder,
 		driveId,
-		parentFolderDataOrID: folderData
+		parentFolderDataOrID
 	}: RecursivePublicBulkUploadParams): Promise<{
 		entityResults: ArFSEntityData[];
 		feeResults: ArFSFees;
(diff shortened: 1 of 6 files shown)


src/ardrive.ts
```diff
@@ -240,6 +241,13 @@ export class ArDrive extends ArDriveAnonymous {
 			throw new Error(errorMessage.cannotMoveIntoSamePlace('File', newParentFolderId));
 		}
 
+		// Assert that there are no duplicate names in the destination folder
+		const entityNamesInParentFolder = await this.arFsDao.getPublicEntityNamesInFolder(newParentFolderId);
+		if (entityNamesInParentFolder.includes(originalFileMetaData.name)) {
+			// TODO: Add optional interactive prompt to resolve name conflicts in ticket PE-599
+			throw new Error(errorMessage.entityNameExists);
+		}
+
```
matibat, line 245: I think we could consolidate this repeated code by extracting these new **async** functions:
```ts
assertUniqueNameInFolder(names: string[], name: string): Promise<void>;
assertNoOverridingFile(files: FileNameAndID[], name: string): Promise<void>;
assertNoOverridingFolder(folders: FolderNameAndID[], name: string): Promise<void>;
```

matibat (commented):
Reviewed by reading. I'll still have to test it.
I'm a little unrelated to bulk uploads, so it was hard to review. I'll carefully test that later on.

Left one comment there

src/ardrive.ts
fedellen: Agreed, and not just the ones from this PR -- we could consolidate a lot of the business logic errors in this class.

I've added a tech_debt ticket for this in phase 1.5: https://ardrive.atlassian.net/browse/PE-609

PR #101, opened 2021-10-19
PE-558 suggestion

matibat:


fedellen (approved):
This is good cleanup, thank you! Much more re-usable

PR #103, opened 2021-10-19
PE-592: Ensure that passwords are used with wallets and not with drive keys

arielmelendez:


src/private_key_data.ts
```diff
@@ -31,6 +31,14 @@ export class PrivateKeyData {
 	private unverifiedDriveKeys: DriveKey[];
 
 	constructor({ password, driveKeys, wallet }: PrivateKeyDataParams) {
+		if (password && !wallet) {
+			throw new Error('Password supplied without a wallet. Did you forget to include your wallet?');
+		}
+
+		if (password && driveKeys) {
+			throw new Error("Password and drive keys can't be used together. Please provide one or the other.");
```
fedellen, line 39: This string is wrapped in double quotes here, I'm surprised prettier format and save and the husky linter hook let that through.

fedellen: Also, I don't think this wrapper in particular would care if both `password && driveKeys` were supplied. It will always try the supplied drive keys first, then fallback to using the password if no driveKeys can decrypt.

PR #107, opened 2021-10-20
PE-563: Derive and assert owner of drive and filter all getters by that owner 

fedellen:
This PR:  

- Derives the owner of a drive by the earliest transaction with that drive-id
- Asserts the provided wallet is consistent with the drive owner in each method that uses a wallet
- Filters all getters by owner of the drive owner

### DIFF
--- a/src/CLICommand/parameters_helper.test.ts
+++ b/src/CLICommand/parameters_helper.test.ts
@@ -250,7 +250,7 @@ describe('ParametersHelper class', () => {
 		it('returns the address of the wallet when a valid --wallet-file is provided', () => {
 			declareCommandWithParams(program, [WalletFileParameter], async (options) => {
 				const parameters = new ParametersHelper(options);
-				expect((await parameters.getWalletAddress()).valueOf()).to.equal(
+				expect(`${await parameters.getWalletAddress()}`).to.equal(
 					'P8aFJizMVBl7HeoRAz2i1dNYkG_KoN7oB9tZpIw6lo4'
 				);
 			});
@@ -260,7 +260,7 @@ describe('ParametersHelper class', () => {
 		it('returns the address of the wallet when a valid --w file is provided', () => {
 			declareCommandWithParams(program, [WalletFileParameter], async (options) => {
 				const parameters = new ParametersHelper(options);
-				expect((await parameters.getWalletAddress()).valueOf()).to.equal(
+				expect(`${await parameters.getWalletAddress()}`).to.equal(
 					'P8aFJizMVBl7HeoRAz2i1dNYkG_KoN7oB9tZpIw6lo4'
 				);
 			});
@@ -272,7 +272,9 @@ describe('ParametersHelper class', () => {
 		// it('returns the address of the wallet when a valid --seed-phrase option is provided', () => {
 		// 	declareCommandWithParams(program, [SeedPhraseParameter], async (options) => {
 		// 		const parameters = new ParametersHelper(options);
-		// expect(await (await parameters.getWalletAddress()).valueOf()).to.equal('P8aFJizMVBl7HeoRAz2i1dNYkG_KoN7oB9tZpIw6lo4');
+		// 		expect(`${await parameters.getWalletAddress()}`).to.equal(
+		//			'P8aFJizMVBl7HeoRAz2i1dNYkG_KoN7oB9tZpIw6lo4'
+		// 		);
 		// 	});
 		// 	CLICommand.parse(program, [
 		// 		...baseArgv,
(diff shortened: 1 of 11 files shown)


tests/integration/ardrive.int.test.ts
```diff
@@ -540,8 +643,9 @@ describe('ArDrive class - integrated', () => {
 				});
 
 				it('throws an error if the file is being moved to a different drive', async () => {
-					stub(arfsDao, 'getPrivateFile').resolves(stubPrivateFile());
-					stub(arDrive, 'getDriveIdAndAssertDrive').resolves(stubEntityIDAlt); // Returns different drive ID
+					stub(arfsDao, 'getPrivateFile').resolves(stubPrivateFile({ driveId: unexpectedDriveId }));
+					stub(arfsDao, 'getOwnerForDriveId').resolves(walletOwner);
+					// stub(arDrive, 'getDriveIdAndAssertDrive').resolves(stubEntityIDAlt); // Returns different drive ID
```
arielmelendez, line 648: Nit: dead?

src/CLICommand/parameters_helper.test.ts
```diff
@@ -268,7 +272,9 @@ describe('ParametersHelper class', () => {
 		// it('returns the address of the wallet when a valid --seed-phrase option is provided', () => {
```
arielmelendez, line 272: What's up with these tests?

fedellen: those are the seed phrase tests we commented out

arielmelendez: Sorry I got lazy. I see that they're the slow --seed-phrase tests

src/ardrive.ts
```diff
@@ -1122,7 +1176,9 @@ export class ArDrive extends ArDriveAnonymous {
 		const createDriveResult = await this.arFsDao.createPublicDrive(
 			driveName,
 			driveRewardSettings,
-			rootFolderRewardSettings
+			rootFolderRewardSettings,
+			// No wallet owner assertion needed on public drive creations
```
arielmelendez, line 1180: Nit: language here is open to interpretation:
`No wallet owner assertion needed` could mean that we need to add an assertion that we don't yet have OR that there's no need to assert it. Since it's the latter, I think it's clearer to say:
`There is no need to assert ownership during drive creation`

src/ardrive.ts
```diff
@@ -1166,7 +1222,9 @@ export class ArDrive extends ArDriveAnonymous {
 			driveName,
 			newPrivateDriveData,
 			driveRewardSettings,
-			rootFolderRewardSettings
+			rootFolderRewardSettings,
+			// Owner has been verified by assertPassword in the CLI layer
```
arielmelendez, line 1226: Will this be sufficient for direct ardrive-core-js integrations though?

src/ardrive.ts
```diff
@@ -1193,11 +1251,6 @@ export class ArDrive extends ArDriveAnonymous {
 		});
 	}
 
-	async getPrivateDrive(driveId: DriveID, driveKey: DriveKey): Promise<ArFSPrivateDrive> {
```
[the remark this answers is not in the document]
arielmelendez: nevermind, I scrolled.

src/ardrive.ts
```diff
@@ -1299,13 +1352,39 @@ export class ArDrive extends ArDriveAnonymous {
 		return { totalPrice: String(totalPrice), totalFilePrice: String(totalFilePrice), communityWinstonTip };
 	}
 
-	async getPrivateFolder(folderId: FolderID, driveKey: DriveKey): Promise<ArFSPrivateFolder> {
-		const folderEntity = await this.arFsDao.getPrivateFolder(folderId, driveKey);
-		return folderEntity;
+	async assertOwnerAddress(owner: ArweaveAddress): Promise<void> {
+		if (`${owner}` !== `${await this.wallet.getAddress()}`) {
```
arielmelendez, line 1356: Javascript/Typescript's object equality stuff gives me a headache, but can we just implement an equality function on ArweaveAddress and use it here? If not possible via a built-in language feature, then something manual like:
```
equalsAddress(other: ArweaveAddress) {
    return this.address === other.address;
}
```

src/commands/list_folder.ts
```diff
@@ -21,7 +21,10 @@ new CLICommand({
 			const driveId = await arDrive.getDriveIdForFolderId(folderId);
 			const driveKey = await parameters.getDriveKey({ driveId });
 
-			children = await arDrive.listPrivateFolder(folderId, driveKey, maxDepth);
+			// We have the drive id from deriving a key, we can derive the owner
+			const driveOwner = await arDrive.getOwnerForDriveId(driveId);
+
+			children = await arDrive.listPrivateFolder(folderId, driveKey, maxDepth, undefined, driveOwner);
```
arielmelendez, line 27: Could benefit from an interface type. Not high priority though.

src/ardrive.ts
fedellen: 8664ba6

src/arweave_address.ts
```diff
@@ -7,6 +7,10 @@ export class ArweaveAddress {
 		}
 	}
 
+	equalsAddress(other: ArweaveAddress): boolean {
+		return this.address === other.address;
+	}
+
```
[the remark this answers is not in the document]
matibat: Ref for chai equality assertion: https://www.chaijs.com/api/bdd/#method_equal

Also tested the https://www.chaijs.com/api/bdd/#method_eql which deeply compares two objects and work just fine for comparing two address instances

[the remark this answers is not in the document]
matibat: Also...
I've found the test passes even when I break them by setting a different valid address. The "expected this to be that" message is logged.

-----------------------------

**EDIT**
Found out the error: the CLICommand class is catching the error so Mocha won't ever realize it

matibat (commented):
When reviewing this PR I've found an error I introduced earlier. It only affects the unhappy path of tests, but we still need to fix it. I'm researching for fixes on this.

----------

I haven't finished reviewing it

PR #122, opened 2021-10-30
PE-635: Safe Types - Winston and AR

arielmelendez:
Started by creating proper Winston and AR types. There are a few tradeoffs involved in this approach:

**Pros:**
• Problems with Winston units are discovered immediately during program input and deserialization from the network
• Strong typing helps create guarantees around value ranges
• Convenience functions reduce reliance on external dependencies
• Capable of computing Winston and AR values in BigNumber ranges

**Cons:**
• Despite its type-safety effectiveness, creating Winston and AR from values is not as syntactically tidy as I'd like. I tried saving some characters by providing the `W` wrapper function around the Winston constructor. There might be some TS magic that can improve this further. I tried looking at BigNumber's implementation of the convenience function for constructing BigNumbers but I couldn't quite get the pattern working in our TS environment. Battle for another day perhaps.
• Arithmetic and mathematical functions need to be treated with care since they're functions now
• Our linear regression library probably can't handle the BigNumber math on its own. But I think there's ways to take its outputs and write a BigNumber wrapper to apply the predictions ourselves in the BigNumber domain.
• Large surface area of changes.
• Need to use string template types to do value equality checks (@matibat maybe you have a better approach?)

**Zero-sum Issues:**
• Dependency on BigNumber library - Silver Lining: We already depended on it transitively through arweave-js
• JSON signature of returned ArDrive objects now prints Winston values as strings. Silver lining - this is more resilient long term as JS JSON parsers won't be able to treat BigNumber values as JSON numbers.

**Suggested PR Sequence:**
1) winston.test.ts
2) winston.ts
3) ar.test.ts
4) ar.ts
5) Everything else


src/types/winston.ts
```diff
@@ -0,0 +1,53 @@
...
+
+	valueOf(): string {
+		return this.amount.toFixed();
+	}
+
+	static max(...winstons: Winston[]): Winston {
+		BigNumber.max();
+		return winstons.reduce((max, next) => (next.amount.isGreaterThan(max.amount) ? next : max));
+	}
+}
```
matibat, line 49: We can maybe use `equals()` for comparison (?)
```suggestion

	static equals(a: Winston, b: Winston): boolean {
		return a.amount.equals(b.amount);
	}
}
```

src/utils/ar_data_price_regression_estimator.ts
```diff
@@ -119,7 +117,14 @@ export class ARDataPriceRegressionEstimator extends AbstractARDataPriceAndCapaci
 		}
 
 		// Return 0 if winston price given does not cover the base winston price for a transaction
-		return Math.max(0, (winston - this.predictor.baseWinstonPrice()) / this.predictor.marginalWinstonPrice());
+		// TODO: Is number sufficient here vs. BigNumber?
```
matibat, line 120: Do you mean for the returned ByteCount?
`MAX_SAFE_INTEGER` bytes to GB = `9007199.254740990698` GB

src/wallet.ts
```diff
@@ -174,7 +179,7 @@ export class WalletDAO {
 			return Promise.resolve({
 				trxID: transaction.id,
 				winston,
-				reward: transaction.reward
+				reward: W(new BigNumber(transaction.reward))
```
matibat, line 182: Can't `W()` take a `string` argument?

matibat (commented):
Left a few comments & questions. The tests are looking awesome!

src/types/winston.ts
arielmelendez: How does this help in the context of Max?

src/utils/ar_data_price_regression_estimator.ts
arielmelendez: Well the regression package uses number rather than BigNumber, so any input data you give it, or outputs it could predict, would have to be in the number range.

src/wallet.ts
arielmelendez: YES! Will update.

src/types.ts
```diff
@@ -28,7 +28,7 @@ export type TransactionID = string; // TODO: make a type that checks lengths
 
 export interface ArDriveCommunityTip {
 	tipPercentage: number;
-	minWinstonFee: number; // TODO: Align with Winston type?
+	minWinstonFee: Winston; // TODO: Align with Winston type?
```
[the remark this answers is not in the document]
arielmelendez: ce5ccea

src/wallet.ts
arielmelendez: 7dbc0f5

src/types/winston.ts
arielmelendez: We don't seem to do any Winston equality checking in the code other than in the tests, and in that context we don't have the agency to call equals directly without eliminating some of the test expectation output beautification we get from chai.

yarn.lock
```diff
@@ -1570,7 +1571,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"bignumber.js@npm:^9.0.1":
+"bignumber.js@npm:^9.0.0, bignumber.js@npm:^9.0.1":
```
matibat, line 1574: This is because some of our deps also depend on bignum, right?

matibat (approved):
It looks good to me!

yarn.lock
arielmelendez: I believe so.

fedellen (approved):
Still not sure about this pattern. It could make parts of our API harder to use or confusing.

There is certainly value added here from input safety and using big number on AR/Winston though. I think it is good to go!

PR #123, opened 2021-11-01
PE-635 Safe EntityID

arielmelendez:
Suggested Reviewing Order:
• entity_id.ts: the class to enforce EntityID format and the EID() convenience function for instantiating EntityIDs.
• entity_id.test.ts: Unit tests
• types.ts: Observe the changes to the aliases. We migrated the existing EntityID, which was a Union Type of the existing entity ID classes, to AnyEntityID since I couldn't just use EntityID in places where a FolderID could be used (since the value might also be a placeholder constant for root folders). This may be something we want to improve upon, e.g. a different constant set for root folders' parent folder IDs, like all 0's.
• arfs_entities.ts: Observe that ArFSEntity, ArFSDriveEntity, and ArFSFileFolderEntity were hoisted out of ardrivecore-js in order to make it them use proper EntityID-aliased classes rather than the string alias that was in types.ts.
    • Also observe that sync_status has been removed
    • Notice that the ArFSDriveEntity has `rootFolderId: FolderID | ENCRYPTED_DATA_PLACEHOLDER_TYPE;` This is because of the constant choice that was made for the SAFE entity types. We may want to improve upon this in the future.


src/types/entity_id.ts
```diff
@@ -0,0 +1,25 @@
...
+	toString(): string {
+		return this.entityId;
+	}
+
+	valueOf(): string {
+		return this.entityId;
+	}
+
+	isEqualTo(entityId: EntityID): boolean {
+		return `${this.entityId}` === `${entityId}`;
```
matibat, line 19: `this.entityId` is already a primitive string
```suggestion
		return this.entityId === `${entityId}`;
```

src/types/entity_id.test.ts
```diff
@@ -0,0 +1,67 @@
...
+			expect(eid2.isEqualTo(eid3), `${eid2} and ${eid3}`).to.be.false;
+			expect(eid3.isEqualTo(eid2), `${eid3} and ${eid2}`).to.be.false;
+		});
+	});
+});
+
+describe('EID function', () => {
+	it('returns the correct EntityID', () => {
+		const expected = new EntityID('01234567-89ab-cdef-0000-000000000000');
+		expect(`${EID('01234567-89ab-cdef-0000-000000000000')}`).to.equal(`${expected}`);
```
matibat, line 65: Maybe use `isEqualTo()` here

matibat: nevermind, that function is already being tested

matibat (commented):
Half-reviewed it

src/types/entity_id.ts
matibat: This is also valid
```suggestion
		return `${this}` === `${entityId}`;
```

src/ardrive.ts
```diff
@@ -232,11 +233,11 @@ export class ArDrive extends ArDriveAnonymous {
 
 		const originalFileMetaData = await this.getPublicFile(fileId);
 
-		if (destFolderDriveId !== originalFileMetaData.driveId) {
+		if (!destFolderDriveId.isEqualTo(originalFileMetaData.driveId)) {
 			throw new Error(errorMessage.cannotMoveToDifferentDrive);
 		}
 
-		if (originalFileMetaData.parentFolderId === newParentFolderId) {
+		if (`${originalFileMetaData.parentFolderId}` === `${newParentFolderId}`) {
```
matibat, line 240: We cannot use `isEqualTo()` because parentFolderId is nullable, right?
```suggestion
		if (originalFileMetaData.parentFolderId.isEqualTo(newParentFolderId)) {
```

src/ardrive.ts
```diff
@@ -290,11 +291,11 @@ export class ArDrive extends ArDriveAnonymous {
 
 		const originalFileMetaData = await this.getPrivateFile(fileId, driveKey);
 
-		if (destFolderDriveId !== originalFileMetaData.driveId) {
+		if (!destFolderDriveId.isEqualTo(originalFileMetaData.driveId)) {
 			throw new Error(errorMessage.cannotMoveToDifferentDrive);
 		}
 
-		if (originalFileMetaData.parentFolderId === newParentFolderId) {
+		if (`${originalFileMetaData.parentFolderId}` === `${newParentFolderId}`) {
```
matibat, line 298: We cannot use `isEqualTo()` because parentFolderId is nullable, right?
```suggestion
		if (originalFileMetaData.parentFolderId.isEqualTo(newParentFolderId)) {
```

src/ardrive.ts
```diff
@@ -344,7 +345,7 @@ export class ArDrive extends ArDriveAnonymous {
 	}
 
 	async movePublicFolder({ folderId, newParentFolderId }: MovePublicFolderParams): Promise<ArFSResult> {
-		if (folderId === newParentFolderId) {
+		if (`${folderId}` === `${newParentFolderId}`) {
```
matibat, line 348:
```suggestion
		if (folderId.isEqualTo(newParentFolderId)) {
```

src/ardrive.ts
```diff
@@ -355,11 +356,11 @@ export class ArDrive extends ArDriveAnonymous {
 
 		const originalFolderMetaData = await this.getPublicFolder(folderId);
 
-		if (destFolderDriveId !== originalFolderMetaData.driveId) {
+		if (!destFolderDriveId.isEqualTo(originalFolderMetaData.driveId)) {
 			throw new Error(errorMessage.cannotMoveToDifferentDrive);
 		}
 
-		if (originalFolderMetaData.parentFolderId === newParentFolderId) {
+		if (`${originalFolderMetaData.parentFolderId}` === `${newParentFolderId}`) {
```
matibat, line 363:
```suggestion
		if (originalFolderMetaData.parentFolderId.isEqualTo(newParentFolderId)) {
```

src/ardrive.ts
```diff
@@ -411,7 +412,7 @@ export class ArDrive extends ArDriveAnonymous {
 	}
 
 	async movePrivateFolder({ folderId, newParentFolderId, driveKey }: MovePrivateFolderParams): Promise<ArFSResult> {
-		if (folderId === newParentFolderId) {
+		if (`${folderId}` === `${newParentFolderId}`) {
```
matibat, line 415:
```suggestion
		if (folderId.isEqualTo(newParentFolderId)) {
```

src/ardrive.ts
```diff
@@ -422,11 +423,11 @@ export class ArDrive extends ArDriveAnonymous {
 
 		const originalFolderMetaData = await this.getPrivateFolder(folderId, driveKey);
 
-		if (destFolderDriveId !== originalFolderMetaData.driveId) {
+		if (!destFolderDriveId.isEqualTo(originalFolderMetaData.driveId)) {
 			throw new Error(errorMessage.cannotMoveToDifferentDrive);
 		}
 
-		if (originalFolderMetaData.parentFolderId === newParentFolderId) {
+		if (`${originalFolderMetaData.parentFolderId}` === `${newParentFolderId}`) {
```
matibat, line 430: Same here, parentFolderId is nullable (?)
```suggestion
		if (originalFolderMetaData.parentFolderId.isEqualTo(newParentFolderId)) {
```

src/utils/arfs_builders/arfs_drive_builders.ts
```diff
@@ -19,11 +19,9 @@ import {
 
 interface DriveMetaDataTransactionData extends EntityMetaDataTransactionData {
 	name: string;
-	rootFolderId: FolderID;
+	rootFolderId: string; // TODO: how can we use real ID types here?
```
matibat, line 22: Wy can't we? :thinking:

matibat (commented):
Left a few suggestions, actually as questions

src/types/entity_id.ts
```diff
@@ -0,0 +1,25 @@
+const entityIdRegex = /^([a-f]|[0-9]){8}-([a-f]|[0-9]){4}-([a-f]|[0-9]){4}-([a-f]|[0-9]){4}-([a-f]|[0-9]){12}$/;
```
matibat, line 1: Are these always uncapitalized? Can we go case `i`nsentitive for hex?
```suggestion
const entityIdRegex = /^([a-f]|[0-9]){8}-([a-f]|[0-9]){4}-([a-f]|[0-9]){4}-([a-f]|[0-9]){4}-([a-f]|[0-9]){12}$/i;
```

matibat: We can poorly simplify this monster regexp, but I think it's worth it :)
```suggestion
const entityIdRegex = /^([a-f\d]){8}-([a-f\d]){4}-([a-f\d]){4}-([a-f\d]){4}-([a-f\d]){12}$/;
```

matibat: We could go even further, but it might lack redability
```suggestion
const entityIdRegex = /^[a-f\d]{8}-([a-f\d]{4}-){3}[a-f\d]{12}$/i;
```

arielmelendez:
Looks like I've broken something:
```
yarn ardrive list-all-drives -w ./CLI_TEST_arweave-keyfile-g1hzNXVbh2M6LMQSUYp7HgkgxdadYqYEfw-HAajlms0.json
Invalid entity ID 'ENCRYPTED'!'
```

src/utils/arfs_builders/arfs_drive_builders.ts
arielmelendez: a112923

src/types/entity_id.ts
arielmelendez: > Are these always uncapitalized? Can we go case `i`nsentitive for hex?

From the Wikipedia entry on UUID:
```
RFC 4122 Section 3 requires that the characters be generated in lower case, while being case-insensitive on input.
```
So I think we can be forgiving on case.

src/arfs_entities.ts
```diff
@@ -1,25 +1,63 @@
...
+
+export const ENCRYPTED_DATA_PLACEHOLDER = 'ENCRYPTED';
+export type ENCRYPTED_DATA_PLACEHOLDER_TYPE = 'ENCRYPTED';
+
+// A Drive is a logical grouping of folders and files. All folders and files must be part of a drive, and reference the Drive ID.
+// When creating a Drive, a corresponding folder must be created as well. This folder will act as the Drive Root Folder.
+// This seperation of drive and folder entity enables features such as folder view queries.
+export interface ArFSDriveEntity extends ArFSEntity {
+	drivePrivacy: string; // identifies if this drive is public or private (and encrypted)  can only be "public" or "private"
+	rootFolderId: FolderID | ENCRYPTED_DATA_PLACEHOLDER_TYPE; // the uuid of the related drive root folder, stored in the JSON data that is uploaded with each Drive Entity metadata transaction
```
fedellen, line 58: Is this necessary here since we added `EncryptedEntityId`?

src/utils/arfs_builders/arfs_drive_builders.ts
```diff
@@ -70,11 +69,11 @@ export class ArFSPublicDriveBuilder extends ArFSMetadataEntityBuilder<ArFSPublic
 			this.appVersion?.length &&
 			this.arFS?.length &&
 			this.contentType?.length &&
-			this.driveId?.length &&
+			this.driveId &&
 			this.entityType?.length &&
 			this.txId?.length &&
 			this.unixTime &&
-			this.driveId == this.entityId &&
+			this.driveId.equals(this.entityId as EntityID) &&
```
fedellen, line 76: Why is this check in the public drive builder but not the private? Do we need this?

src/types.ts
```diff
@@ -10,10 +11,11 @@ export type SeedPhrase = string;
 
 export type NetworkReward = Winston;
 
-export type FolderID = string;
-export type FileID = string;
-export type DriveID = string;
-export type EntityID = DriveID | FolderID | FileID;
+export type RootFolderID = 'root folder';
+export type FolderID = EntityID | RootFolderID;
```
fedellen, line 15: I don't think this is technically correct here. And this is also causing the type issues downstream where you have to cast things `as EntityID` in a couple places in the PR.

A folder's `EntityID` will never be `'root folder'`. That string is only ever used as a place holder for the `parentFolderId` field when it is missing. This case should only happen on the root folder of a drive. So, `folderId` will never be root folder, only the `parentFolderId` field.

```suggestion
export type FolderID = EntityID;
```

arielmelendez: Yep those placeholder values were causing exceptions at runtime. I'm going to try the same pattern we used for EncryptedEntityId.

src/arfs_entities.ts
arielmelendez: 2738016f1886bf5b426e20a42adfbad322765fd6

src/utils/arfs_builders/arfs_drive_builders.ts
arielmelendez: befad39

src/types.ts
arielmelendez: 2738016f1886bf5b426e20a42adfbad322765fd6

src/types/entity_id.ts
arielmelendez: c90ae71

Good stuff @matibat !

src/ardrive.ts
matibat: parent folders are nullable

src/arfsdao.ts
```diff
@@ -744,7 +742,7 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		while (hasNextPage) {
 			const gqlQuery = buildQuery({
 				tags: [
-					{ name: 'Parent-Folder-Id', value: folderIDs },
+					{ name: 'Parent-Folder-Id', value: folderIDs.map(toString) },
```
matibat, line 745: is this `toString`:
 - window.toString
 - this.toString
 - Object.prototype.toString
 - none of them (?
 
 let's maybe do the mapping with an arrow function

src/utils/arfs_builders/arfs_drive_builders.ts
```diff
@@ -210,6 +210,14 @@ export class ArFSPrivateDriveBuilder extends ArFSMetadataEntityBuilder<ArFSPriva
 	}
 }
 
+// A utility type to assist with fail-safe decryption of private entities
+export class EncryptedEntityID extends EntityID {
+	constructor() {
+		super(`${stubEntityID}`); // Unused after next line
+		this.entityId = ENCRYPTED_DATA_PLACEHOLDER;
```
matibat, line 217: So any encrypted entity id will equal any other unrelated encrypted entity id? :thinking:

matibat: Does that comparison happen somewhere?

src/utils/arfs_builders/arfs_folder_builders.ts
```diff
@@ -4,13 +4,25 @@ import { ArFSPrivateFolder, ArFSPublicFolder } from '../../arfs_entities';
...
+import { EID, EntityID } from '../../types/entity_id';
+import { stubEntityID } from '../stubs';
+
+export const ROOT_FOLDER_ID_PLACEHOLDER = 'root folder';
+
+// A utility type to assist with fail-safe decryption of private entities
+export class RootFolderID extends EntityID {
+	constructor() {
+		super(`${stubEntityID}`); // Unused after next line
+		this.entityId = ROOT_FOLDER_ID_PLACEHOLDER;
```
matibat, line 16: Beware all root folder ids will equal for any drive

matibat (changes requested):
It's looking great! I'm happy with the implementation of `toPrimitive`.

I left a pair of questions on which I see a possible risk, but maybe there isn't. I'll approve after those last comments are addressed :+1: 
_requesting changes because of the `map(toString)` which I think it actually affects that query_

src/ardrive.ts
matibat: exactly.

src/utils/arfs_builders/arfs_folder_builders.ts
arielmelendez: We never query against this. It's just for display purposes. But yes it's sketchy and I don't love it.

src/utils/arfs_builders/arfs_drive_builders.ts
arielmelendez: Similar to root folder this is just for display purposes for "fail safe decryption" of private entities for the list-all-drives command.

src/arfsdao.ts
arielmelendez: Let's not leave it to chance 😬 2663b2b

src/utils/arfs_builders/arfs_folder_builders.ts
```diff
@@ -4,13 +4,25 @@ import { ArFSPrivateFolder, ArFSPublicFolder } from '../../arfs_entities';
 import { ArweaveAddress } from '../../types/arweave_address';
 import { CipherIV, DriveKey, FolderID } from '../../types';
 import { ArFSFileOrFolderBuilder } from './arfs_builders';
+import { EID, EntityID } from '../../types/entity_id';
+import { stubEntityID } from '../stubs';
+
+export const ROOT_FOLDER_ID_PLACEHOLDER = 'root folder';
+
+// A utility type to assist with fail-safe decryption of private entities
```
fedellen, line 12: This would be a utility type that creates a placeholder for the parentFolderId field when its missing; which presumably means this folder is the root folder of the drive.

src/ardrive.ts
matibat: I'm fine with the current change

src/ardrive.ts
matibat: I'm fine with the current change

matibat (approved):
Thanks for your fixes! LGTM now!

src/utils/arfs_builders/arfs_folder_builders.ts
arielmelendez: yea lemme update this

src/types/entity_id.ts
matibat: This is redundant, but it's not the end of the world

src/utils/arfs_builders/arfs_folder_builders.ts
arielmelendez: 8bca2e9

fedellen (approved):

PR #126, opened 2021-11-01
PE-633: Uploads will "upsert" by default

fedellen:
This PR adds the `--upsert` parameter which will become the new default on single and bulk file uploads. 

On a FILE to FILE name conflict within the destination folder, the upsert option will compare `lastModifiedDate` on the files. If they match, the upload will be skipped. Otherwise a new REVISION will be created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,8 +707,8 @@ Expect the behaviors from the following table for each resolution setting:
 | Source Type | Conflict at Dest | `skip` | `replace` | `upsert` (default) |
 | ----------- | ---------------- | ------ | --------- | ------------------ |
 | File        | None             | Insert | Insert    | Insert             |
-| File        | Matching File    | Skip   | Update    | skip               |
-| File        | Different File   | Skip   | Update    | update             |
+| File        | Matching File    | Skip   | Update    | Skip               |
+| File        | Different File   | Skip   | Update    | Update             |
 | File        | Folder           | Skip   | Fail      | Fail               |
 | Folder      | None             | Insert | Insert    | Insert             |
 | Folder      | File             | Skip   | Fail      | Fail               |
--- a/README.md
+++ b/README.md
@@ -690,7 +690,9 @@ NOTE: To upload to the root of a drive, specify its root folder ID as the parent
 ardrive drive-info -d "c7f87712-b54e-4491-bc96-1c5fa7b1da50" | jq -r '.rootFolderId'
 ```
 
-By default, the single `upload-file` command will skip name conflicts found. To override this behavior and make a new revision of a file, use the `--replace` option:
+By default, the single `upload-file` command will `--upsert` on name conflicts found. This means that when it finds a file in the destination folder with the same name, it will compare the last modified dates of that file and the file to upload. If they are matching, the upload will be skipped, otherwise the upload will be added as a new revision.
+
+To override this behavior, use the `--replace` option to always make new revisions of a file or the `--skip` option to always skip the upload on name conflicts:
 
 ```shell
 ardrive upload-file --replace --local-file-path /path/to/file.txt  --parent-folder-id "9af694f6-4cfc-4eee-88a8-1b02704760c0" -w /path/to/wallet.json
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -690,7 +690,9 @@ NOTE: To upload to the root of a drive, specify its root folder ID as the parent
 ardrive drive-info -d "c7f87712-b54e-4491-bc96-1c5fa7b1da50" | jq -r '.rootFolderId'
 ```
 
-By default, the single `upload-file` command will skip name conflicts found. To override this behavior and make a new revision of a file, use the `--replace` option:
+By default, the single `upload-file` command will `--upsert` on name conflicts found. This means that when it finds a file in the destination folder with the same name, it will compare the last modified dates of that file and the file to upload. If they are matching, the upload will be skipped, otherwise the upload will be added as a new revision.
```
arielmelendez, line 693: Don't forget to mention the "insert" part of "upsert"

README.md
```diff
@@ -709,9 +711,9 @@ This method of upload can be used to upload a large number of files and folders
 -   Folder names that conflict with a FILE name at the destination will cause an error to be thrown
 -   Folder names that conflict with a FOLDER name at the destination will use the existing folder ID (i.e. skip) rather than creating a new folder
 -   File names that conflict with a FOLDER name at the destination will cause an error to be thrown
--   File names that conflict with a FILE name at the destination will be SKIPPED
+-   When the intended file name conflicts with a FILE name at the destination the file will be SKIPPED if the last modified date matches the file to upload. If they have different last modified dates, it will be uploaded as a new REVISION
```
arielmelendez, line 714: Does it make sense to mention how this varies by flags provided?

README.md
```diff
@@ -709,9 +711,9 @@ This method of upload can be used to upload a large number of files and folders
...
 
-Similar to the single file upload, the above FILE to FILE name conflict resolution behavior can be modified by the `--replace` command. This will force new revisions on all conflicts within the bulk upload.
+Similar to the single file upload, the above FILE to FILE name conflict resolution behavior can be modified. Use the `--replace` option to will force new revisions on all conflicts within the bulk upload regardless of last modified date. Or use the `--skip` option to simply skip all FILE to FILE name conflicts.
```
arielmelendez, line 716: Grammar problem here

arielmelendez: > to simply skip all FILE to FILE name conflicts.
Will that not apply to FILE/FOLDER name conflicts as well?

src/ardrive.ts
```diff
@@ -528,10 +528,22 @@ export class ArDrive extends ArDriveAnonymous {
 
 		const conflictingFileName = filesAndFolderNames.files.find((f) => f.fileName === destFileName);
 
-		if (conflictingFileName && conflictResolution === 'skip') {
-			return emptyArFSResult;
+		if (conflictingFileName) {
+			if (conflictResolution === 'skip') {
```
arielmelendez, line 532: Can we get some constants for these flag values?

tests/integration/ardrive.int.test.ts
```diff
@@ -451,15 +451,24 @@ describe('ArDrive class - integrated', () => {
 		});
 
 		describe('file function', () => {
+			const matchingLastModifiedDate = 420;
```
arielmelendez, line 454: Mah man

arielmelendez (changes requested):
Generally looks good! Just a few documentation issues that need updating.

README.md
fedellen: I've reworked this section, let me know what you think!

README.md
fedellen: Good catch! 

I've added that functionality in the latest commits. The single file update was pretty straightforward.

When adding this for the bulk side, I took the opportunity to refactor the `checkAndAssignNames` function out of the ArDrive class and into the folder wrapper class. There was nothing keeping it there and I think the method makes more sense within the folder scope.

In doing this, I lifted all of the error throwing into the recursive method so I could properly `skip` them if the conflict resolution setting was appropriate.

src/ardrive.ts
fedellen: Added: c1f8da0

src/ardrive.ts
```diff
@@ -1049,7 +1000,14 @@ export class ArDrive extends ArDriveAnonymous {
 		let uploadEntityResults: ArFSEntityData[] = [];
 		let folderId: FolderID;
 
-		if (wrappedFolder.existingId) {
+		if (wrappedFolder.fileNameConflict) {
+			if (conflictResolution === skipOnConflicts) {
+				// Return empty result on skip
+				return { entityResults: uploadEntityResults, feeResults: uploadEntityFees };
```
arielmelendez, line 1006: You might want to return actual empties here just in case someone changes something upstream later on.

src/arfs_file_wrapper.ts
```diff
@@ -127,6 +130,70 @@ export class ArFSFolderToUpload {
...
+			// Folder has no existing ID to check
+			return;
+		}
+
+		const existingEntityNamesAndIds = await getExistingNamesFn(this.existingId);
+
+		for await (const file of this.files) {
+			const baseFileName = file.getBaseFileName();
+
+			const folderNameConflict = existingEntityNamesAndIds.folders.find(
```
arielmelendez, line 146: One note on the name choice: It's not clear from the variable name whether `folder` in `folderNameConflict` is the source folder or the destination folder. Consider `existingFolderAtDestConflict`. Similar concept can be applied to all other conflicts.

arielmelendez (approved):
Looks good. The only other changes I'd consider other than those I commented on above might be a concise table in the README that rolls up all the conflict resolution strategies in a simple way. Something like:
```
src type | conflict at dest | skip | replace | upsert
file | none | insert | insert | insert
file | same file | skip |  update | skip
updated file | old file | skip | update | update
file | folder | skip | fail | fail
folder | none | insert | insert | insert
folder | file | skip | fail | fail
folder | folder | skip | reuse | reuse.  <-- Double check this last one with team!
```

arielmelendez (approved):

PR #129, opened 2021-11-02
PE-635: Safe TransactionID type

arielmelendez:


src/types/transaction_id.ts
```diff
@@ -0,0 +1,23 @@
+const trxIdRegex = /^([a-zA-Z]|[0-9]|-|_){43}$/;
```
matibat, line 1: It can be simplified
```suggestion
const trxIdRegex = /^([a-zA-Z0-9]|-|_){43}$/;
```

matibat: Also, confirm this: are underscores allowed?

matibat: `\w` is the same than `[a-zA-Z0-9]|_`
```suggestion
const trxIdRegex = /^(\w|-){43}$/;
```

src/types/transaction_id.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class TransactionID {
+	constructor(private readonly transactionId: string) {
+		if (!transactionId.match(trxIdRegex)) {
+			throw new Error(
+				'Transaction ID should be a 43-character, alphanumeric string potentially including "=" and "_" characters.'
```
matibat, line 7: > potentially including "="

Because it's a base64. Where do we trim the trailing equals?

matibat: Should we accept it in the regex?

matibat (commented):
Review in progress.
Left a comment for simplification and a few questions

src/arfs_entity_result_factory.ts
```diff
@@ -1,5 +1,6 @@
 import { ArFSFileMetadataTransactionData } from './arfs_trx_data_types';
-import { TransactionID, DriveID, FolderID, FileID, FileKey, DriveKey } from './types';
+import { DriveID, FolderID, FileID, FileKey, DriveKey } from './types';
+import { TransactionID } from './types/transaction_id';
```
matibat, line 3: Can't we export the safe types from `./types/index.ts`?

matibat (commented):
I'm so happy to see these new types! :rocket:

src/types/transaction_id.ts
arielmelendez: I am following up on what number domain trx IDs are in.

src/types/transaction_id.ts
arielmelendez: 46f46fe

src/types/transaction_id.ts
arielmelendez: Based on convos with Arweave team in Discord it seems that =, though a valid Base64 URL character, would only appear at the end of a hashed string, and only if padding were required. Apparently no padding is required for the hashes we create here.

src/arfs_entity_result_factory.ts
arielmelendez: 3824924
Exported as a folder.

src/types/transaction_id.ts
matibat: We are covered here. The network doesn't require the trailing equals

src/arfs_entity_result_factory.ts
matibat: It looks better! **-** We are having conflicts importing between `src/types.ts` and `src/types/`. Looking forward to that change

matibat: Maybe I'd expect that change at #130 or further PRs for PE-635

matibat (approved):
It's looking good to me!
Just looking forward to the upcoming `types` import fix

src/arfs_entity_result_factory.ts
arielmelendez: Yea I'll take care of it for ByteCount

fedellen (approved):
👍

PR #130, opened 2021-11-03
PE-635: Safer SeedPhrase type

arielmelendez:


src/types/seed_phrase.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	toString(): string {
+		return this.seedPhrase;
+	}
+
+	valueOf(): string {
+		return this.seedPhrase;
+	}
+
+	toJSON(): string {
```
matibat, line 28: Is this function in use? If so, please add a docblock giving some context

src/types/seed_phrase.ts
```diff
@@ -0,0 +1,35 @@
+import { Equatable } from './equatable';
+
+const seedPhraseRegex = /^(\b[a-zA-Z]+\b(\s+(?=\b[a-zA-Z]+\b)|$)){12}$/;
```
matibat, line 3: To-Do for @matibat: review this regexp

src/types.ts
```diff
@@ -6,7 +6,6 @@ export const DEFAULT_APP_NAME = 'ArDrive-Core';
 export const DEFAULT_APP_VERSION = '1.0.0';
 
 export type PublicKey = string;
-export type SeedPhrase = string;
```
matibat, line 9: The easiest way to solve the import issue is:
 - Move this source file inside `src/types/`
 - Add this line to `src/types/index.ts`: `export * from './types.ts'`

The import tree will remain the same until we finish the types refactor when we'll surely delete/rename that file

matibat (approved):
Very atomic PR. It was super easy to review! :partying_face: 

Approved with some minor comemnts

src/types/seed_phrase.ts
matibat: This method is used by `JSON.stringify` internally, that's why there's no references to it in the codebase

fedellen (approved):

src/types/seed_phrase.ts
matibat: Okay, I now see the point of that positive look-ahead.
1. It has to be a look-ahead to exclude it from the match, and
2. it has to match the next word or the end of the line.
3. Then because of 1. the space is outside the look-ahead because we want to match it.
The purpose is to avoid the trailing space. It would be kind of: "match all words that are followed by more than one space (included) and another word, or the end of the line"

Now, see this:
```suggestion
const seedPhraseRegex = /^(\b[a-zA-Z]+\b(\s+\b|$)){12}$/;
```
The phrase for this regexp would be: "match all words that are followed by more than one space and a word boundary (both included), or the end of the line"
With that regexp I took advantage of `\b` not being an actual character so we can get rid of the look-ahead

matibat: We can simplify a little more by adding the case `i`nsentitive flag
```suggestion
const seedPhraseRegex =/^(\b[a-z]+\b(\s+\b|$)){12}$/i;
```
although I'm not sure if the seed phrase accepts capitalized characters

matibat: @arielmelendez what dya think?

PR #131, opened 2021-11-03
PE-652: list methods produce full file entity output

fedellen:
This PR extends a workaround to display all data json fields on the `list-folder` and `list-drive` commands.


src/arfs_entities.ts
```diff
@@ -245,6 +256,8 @@ export class ArFSPublicFolder extends ArFSFileOrFolderEntity {
 			txId,
 			unixTime,
 			0,
+			stubTransactionID,
+			'application/json',
```
arielmelendez, line 260: Wouldn't hate a constant for this. Can totally see us misspelling it somewhere at some point.

arielmelendez (dismissed):

src/arfs_prototypes.ts
```diff
@@ -210,7 +210,7 @@ export class ArFSPrivateFileMetaDataPrototype extends ArFSFileMetaDataPrototype
 
 export abstract class ArFSFileDataPrototype extends ArFSObjectMetadataPrototype {
 	abstract readonly objectData: ArFSFileDataTransactionData;
-	abstract readonly contentType: DataContentType | 'application/octet-stream';
+	abstract readonly contentType: DataContentType | typeof PRIVATE_CONTENT_TYPE;
```
arielmelendez, line 213: Will this typeof declaration ensure that the only other value here is 'application/octet-stream' or will it allow any string?

arielmelendez (approved):

src/arfs_prototypes.ts
fedellen: Using typeof will constrain the type to 'application/octet-stream'. I don't think this piece of code is a good example of that though. Here, `DataContentType` is already a string, so the type will be `string | 'application/octet-stream'`. Which basically just means string.

Here is a better example that shows the behavior of `typeof` on a string const declaration:

```ts
export const PRIVATE_CONTENT_TYPE = 'application/octet-stream';

export type privContentType = typeof PRIVATE_CONTENT_TYPE;

const myContentType: privContentType = 'text/plain'; // ❌ Error
// Type '"text/plain"' is not assignable to type '"application/octet-stream"'.ts(2322)

const myPrivateContentType: privContentType = 'application/octet-stream'; // 👍

```

fedellen: Actually.. this just shouldnt be `DataContentType` here at all. These classes only deal with the meta data entity. They don't use data content type, only content type: which is only `application/json` or `application/octet-stream` afaik

I'll clean that up here

matibat: In conclusion... This is gonna be not an issue when we have a real DataContentType, not aliasing string. Right?

matibat (approved):
It looks good to me!
Also working on my end for the `list-folder` on a private drive

PR #132, opened 2021-11-03
PE-635: Safer FeeMultiple 

arielmelendez:


src/types/fee_multiple.ts
```diff
@@ -0,0 +1,42 @@
...
+	toString(): string {
+		return `${this.feeMultiple}`;
+	}
+
+	valueOf(): number {
+		return this.feeMultiple;
+	}
+
+	equals(other: FeeMultiple): boolean {
+		// Hard to say what precision to use here. Use EPSILON by default since most multiples will be pretty small
```
matibat, line 27: We actually don't use it in the codebase at all. Also, `FeeMultiple` doesn't implement `Equatable<T>`
Maybe we don't need this method in this specific type (?

src/types/fee_multiple.ts
```diff
@@ -0,0 +1,42 @@
+import BigNumber from 'bignumber.js';
+
+export class FeeMultiple {
+	constructor(private readonly feeMultiple: number) {
+		if (this.feeMultiple < 1.0 || Number.isNaN(feeMultiple) || !Number.isFinite(feeMultiple)) {
```
matibat, line 5: I think we don't need to check for the maximum. There's no maximum boost, or there is it?

matibat: @matibat: the theorical max is MAX_SAFE_INTEGER for now
```ts
!Number.isFinite(feeMultiple)
```
**Edit** See the test case of "can boost large rewards"

matibat: _and the wallet balance_ :rofl:

src/wallet.ts
```diff
@@ -118,9 +118,9 @@ export class WalletDAO {
 			trxAttributes.last_tx = 'STUB';
 		}
 		const transaction = await this.arweave.createTransaction(trxAttributes, jwkWallet.getPrivateKey());
-		if (rewardSettings.feeMultiple && rewardSettings.feeMultiple > 1.0) {
+		if (rewardSettings.feeMultiple?.wouldBoostReward()) {
 			// Round up with ceil because fractional Winston will cause an Arweave API failure
```
matibat, line 122: Update old comment

src/arfsdao.ts
```diff
@@ -649,16 +649,16 @@ export class ArFSDAO extends ArFSDAOAnonymous {
 		const transaction = await this.arweave.createTransaction(trxAttributes, wallet.getPrivateKey());
 
 		// If we've opted to boost the transaction, do so now
-		if (rewardSettings.feeMultiple && rewardSettings.feeMultiple > 1.0) {
+		if (rewardSettings.feeMultiple?.wouldBoostReward()) {
 			// Round up with ceil because fractional Winston will cause an Arweave API failure
```
matibat, line 653: Update old comment

src/CLICommand/parameters_helper.ts
```diff
@@ -10,14 +10,16 @@ import {
...
 import { cliWalletDao } from '..';
 import { DriveID, DriveKey } from '../types';
 import passwordPrompt from 'prompts';
 import { PrivateKeyData } from '../private_key_data';
 import { ArweaveAddress } from '../types/arweave_address';
 import { SeedPhrase } from '../types/seed_phrase';
+import { FeeMultiple } from '../types/';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type ParameterOptions = any;
```
matibat, line 25: To-Do for @matibat: a dict is more specific than `any`

```suggestion
type ParameterOptions = Register<string, unknown>;
```

matibat: And when there is a `ParameterName` type, I'd suggest:
```suggestion
type ParameterOptions = Register<ParameterName, unknown>;
```

src/types/ar.ts
```diff
@@ -15,7 +15,7 @@ export class AR {
 
 	toString(): string {
 		BigNumber.config({ DECIMAL_PLACES: 12 });
-		const w = new BigNumber(this.winston.toString());
+		const w = new BigNumber(this.winston.toString(), 10);
```
matibat, line 18: Why is that?

matibat (changes requested):
It's looking good! Minor changes requested

src/types/ar.ts
arielmelendez: It specifies the base number system to use when considering the config setting. It might be superfluous since tests were passing before.

src/wallet.ts
arielmelendez: 3821ad7

Thanks!

src/arfsdao.ts
arielmelendez: 3821ad7

Thanks!

src/types/fee_multiple.ts
arielmelendez: I'm inclined to agree. I'll remove it.

fedellen (approved):
Good to go from my end!

src/types/fee_multiple.ts
arielmelendez: 54285dd

src/types/ar.ts
matibat: Got ya

matibat (approved):

PR #134, opened 2021-11-04
PE-635: safe unix time

arielmelendez:


src/types/unix_time.ts
```diff
@@ -0,0 +1,27 @@
...
+			throw new Error('Unix time must be a positive integer!');
+		}
+	}
+
+	[Symbol.toPrimitive](hint?: string): number | string {
+		if (hint === 'string') {
+			this.toString();
+		}
+
+		return this.unixTime;
```
matibat, line 13: For each of these `toPrimitive` we have to carefully think what are we gonna return for the `"default"` case
I think for most of the cases we are covered. Let's review this when all branches are mreged into the integration branch

matibat, line 13: For each of these `toPrimitive` we have to carefully think what are we gonna return for the `"default"` case
I think for most of the cases we are covered. Let's review this when all branches are mreged into the integration branch

matibat, line 13: For each of these `toPrimitive` we have to carefully think what are we gonna return for the `"default"` case
I think for most of the cases we are covered. Let's review this when all branches are mreged into the integration branch

matibat: We ain't covering this case in the tests neither

src/types/unix_time.ts
```diff
@@ -0,0 +1,27 @@
+export class UnixTime {
+	constructor(private readonly unixTime: number) {
+		if (this.unixTime < 0 || !Number.isInteger(this.unixTime) || !Number.isFinite(this.unixTime)) {
+			throw new Error('Unix time must be a positive integer!');
+		}
```
matibat, line 5: From `src/arfs_entities.ts`:
```ts
export class ArFSEntity {
	...
	unixTime: UnixTime; // seconds since unix epoch, taken at the time of upload, 10 numbers eg. 1620068042
	...
}
```

```suggestion
		if (this.unixTime < 0 || !Number.isInteger(this.unixTime) || !Number.isFinite(this.unixTime)) {
			throw new Error('Unix time must be a positive integer!');
		}
		const length = this.toString().length;
		if (length !== 10) {
				throw new Error(`Unix time must be 10 digits long, got ${length}`);
		}
```

matibat (changes requested):
Requesting changes for the unix time length only. Let's see if adding that check does worth it

src/types/unix_time.ts
matibat: Bug on github: When you multiple-tap `add review commen` it adds as may as you tapped ahha

matibat: Bug on github: When you multiple-tap `add review commen` it adds as may as you tapped ahha

src/types/unix_time.ts
arielmelendez: So if we do this, we'd need separate types for the two number domains for lastModifiedDate and unixTime...

src/types/unix_time.ts
arielmelendez: I thought about that a bit and in all those cases I think we're generally concerned with the raw underlying value. The object wrappers are just containers for code.

src/types/unix_time.ts
matibat: I see... Yeah, we are fine as it is :+1:

matibat (approved):

fedellen (approved):
🙌

PR #138, opened 2021-11-04
PE-635 Safe Types

arielmelendez:
This is the integration branch of all the individual safe type branches.


src/utils/ardrive.test.ts
```diff
@@ -171,9 +158,8 @@ describe('ArDrive class', () => {
 			});
 
 			const actual = await arDrive.estimateAndAssertCostOfFolderUpload(stubPublicFileTransactionData);
-			expect(actual).to.deep.equal({
-				metaDataBaseReward: '147'
-			});
+			// TODO: Bummer to lose deep equal verification
```
matibat, line 161: How so? I didn't get this comment

src/utils/arfs_builders/arfs_file_builders.ts
```diff
@@ -91,7 +100,7 @@ export class ArFSPublicFileBuilder extends ArFSFileBuilder<ArFSPublicFile> {
 					this.txId,
 					this.unixTime,
 					this.parentFolderId,
-					this.entityId,
+					this.entityId as FileID,
```
matibat, line 103: Why this casting here?

src/utils/arfs_builders/arfs_file_builders.ts
```diff
@@ -199,7 +208,7 @@ export class ArFSPrivateFileBuilder extends ArFSFileBuilder<ArFSPrivateFile> {
 				this.txId,
 				this.unixTime,
 				this.parentFolderId,
-				this.entityId,
+				this.entityId as FileID,
```
matibat, line 211: Double check this casting here

src/utils/data_price_regression.ts
```diff
@@ -21,7 +21,8 @@ export class ARDataPriceRegression {
 		}
 
 		const dataPoints: DataPoint[] = pricingData.map(
-			(pricingDatapoint) => [pricingDatapoint.numBytes, pricingDatapoint.winstonPrice] as DataPoint
+			// TODO: BigNumber regressions
```
matibat, line 24: Do we need a tech_debt ticket?

src/utils/mapper_functions.ts
```diff
@@ -25,5 +25,5 @@ export function folderToNameAndIdMap(entity: ArFSFileOrFolderEntity): FolderName
 }
 
 export function fileToNameAndIdMap(entity: ArFSFileOrFolderEntity): FileNameAndId {
-	return { fileId: entity.entityId, fileName: entity.name };
+	return { fileId: entity.entityId as FileID, fileName: entity.name };
```
matibat, line 28: Is this casting fine?

tests/integration/ardrive.int.test.ts
```diff
@@ -25,18 +25,20 @@ import { expectAsyncErrorThrow } from '../../src/utils/test_helpers';
...
-import { DriveKey, FileID, Winston } from '../../src/types';
+import { DriveKey, FileID, W, Winston, FeeMultiple, EID } from '../../src/types';
 import { ArFSFileToUpload, wrapFileOrFolder } from '../../src/arfs_file_wrapper';
+import { RootFolderID } from '../../src/utils/arfs_builders/arfs_folder_builders';
 
-const entityIdRegex = /^([a-f]|[0-9]){8}-([a-f]|[0-9]){4}-([a-f]|[0-9]){4}-([a-f]|[0-9]){4}-([a-f]|[0-9]){12}$/;
+// Don't use the existing constants just to make sure our expectations don't change
+const entityIdRegex = /^[a-f\d]{8}-([a-f\d]{4}-){3}[a-f\d]{12}$/i;
 const trxIdRegex = /^([a-zA-Z]|[0-9]|-|_){43}$/;
 const fileKeyRegex = /^([a-zA-Z]|[0-9]|-|_|\/|\+){43}$/;
```
matibat, line 35:
```ts
// Don't use the existing constants just to make sure our expectations don't change
```
But shouldn't these be an exact copy of the original ones?

matibat (commented):
Seems good to go!

I have some concerns on which I commented. After checking that I'll approve

src/CLICommand/parameters_helper.ts
```diff
@@ -216,14 +221,21 @@ export class ParametersHelper {
...
+	public getRequiredParameterValue<T = string>(
+		parameterName: ParameterName,
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		mapFunc: (input: any) => T = (input: any) => input as T
+	): T {
+		// FIXME: it could also return an array or a boolean!
 		const value = this.options[parameterName];
 		if (!value) {
 			throw new Error(`Required parameter ${parameterName} wasn't provided!`);
 		}
```
fedellen, line 238: Shouldn't this work without the anys? 

```ts
	public getRequiredParameterValue<T = string>(
		parameterName: ParameterName,
		mapFunc: (value: T) => T = (value: T) => value as T
	): T {
		const value = this.options[parameterName];
		if (value === undefined) {
			throw new Error(`Required parameter ${parameterName} wasn't provided!`);
		}
		return mapFunc(value);
```

And I don't see why boolean/arrays aren't okay here, we have a generic. We should be able to call this on a required boolean like so:

```ts
const dryRun = parameters.getRequiresParameterValue<boolean>(DryRunParameter);
```

I might be missing something with how they're returned by commander though

arielmelendez: The input value isn't a T. It's whatever options.whatever is, and they are seen by the compiler as `any`

tests/integration/ardrive.int.test.ts
arielmelendez: Yes these have migrated apart. We should unify, but it's not critical for this branch.

src/utils/mapper_functions.ts
arielmelendez: Appears to be unnecessary any more. I can excise all examples of this.

src/utils/data_price_regression.ts
arielmelendez: Probably.

src/utils/arfs_builders/arfs_file_builders.ts
arielmelendez: Unnecessary. Will excise.

src/utils/ardrive.test.ts
arielmelendez: Stale comment. Will remove.

fedellen:
Getting two "hidden" errors during the `yarn test` execution:

```shell
Error: Error from inside the action
error: required option '-p --required-type-parameter <requiredTypeParameter>' not specified
```

Not critical, but I'm a bit perplexed why we still have errors hidden from mocha

src/utils/mapper_functions.ts
arielmelendez: 91faa5f

src/CLICommand/parameters_helper.ts
fedellen: The options type in this file doesn't match the options type from the action class. It should have the type from `export type ParsedParameters = Record<ParameterName, unknown>;`

And when we add the generic, we know exactly what unknown is (or should be). So the options in this function would be `Record<ParameterName, T>` rather than any.

I think it could be easily represented by casting `const value: T = this.options[parameterName]`

src/commands/create_folder.ts
```diff
@@ -20,19 +21,20 @@ new CLICommand({
 		DryRunParameter,
 		...DrivePrivacyParameters
 	],
-	async action(options) {
+	action: new CLIAction(async function action(options) {
 		const parameters = new ParametersHelper(options);
 		const wallet: Wallet = await parameters.getRequiredWallet();
+		const dryRun = !!parameters.getParameterValue(DryRunParameter);
```
fedellen, line 27: Any reason we didn't extend the generic parameter pattern to this parameter getter? 

```suggestion
              const dryRun = parameters.getParameterValue<boolean>(DryRunParameter);
```

src/commands/send_ar.ts
```diff
@@ -1,38 +1,39 @@
...
 	name: 'send-ar',
 	parameters: [ArAmountParameter, DestinationAddressParameter, WalletFileParameter, BoostParameter, DryRunParameter],
-	async action(options) {
-		assertARPrecision(options.arAmount);
+	action: new CLIAction(async function action(options) {
 		const parameters = new ParametersHelper(options);
-		const arAmount: number = +options.arAmount;
-		const destAddress = new ArweaveAddress(options.destAddress);
+		const arAmount = parameters.getRequiredParameterValue(ArAmountParameter, AR.from);
+		const destAddress = parameters.getRequiredParameterValue(DestinationAddressParameter, (addr) => ADDR(addr));
```
fedellen, line 21:
```suggestion
		const destAddress = parameters.getRequiredParameterValue(DestinationAddressParameter, ADDR);
```

src/utils/filter_methods.ts
```diff
@@ -34,7 +33,7 @@ export function latestRevisionFilterForDrives(
 	_index: number,
 	allEntities: ArFSDriveEntity[]
 ): boolean {
-	const allRevisions = allEntities.filter((e) => e.driveId === entity.driveId);
+	const allRevisions = allEntities.filter((e) => e.driveId.equals(entity.driveId));
 	const latestRevision = allRevisions[0];
 	return entity.txId === latestRevision.txId;
```
fedellen, line 38: Shouldn't this be:

```suggestion
	return entity.txId.equals(latestRevision.txId);
```

fedellen (commented):
Left some comments. The only one that looks critical is that `txId` comparison using `===` rather than the new `equals` method

src/utils/ardrive.test.ts
arielmelendez: Actually it's not stale. Equality on the safe types is not as nice to test against. You either need to do:
```
expect(objA.equals(new SafeTypeThing('value i care about')).to.be.true;
```
or do as above and cast to a reduced value (e.g. 'value i care about', i.e. '147') to use the chai expectation functions.

tests/integration/ardrive.int.test.ts
arielmelendez: 3cb7221

src/commands/create_folder.ts
arielmelendez: Well there's something else going on here beyond type mapping which is that we're also choosing a default value in the absence of it existing. e.g. `!!undefined` is `false`. I do think we could improve getParameterValue to do type mapping, but it should also provide for supplying a default value in the absence of a set option. I wanted to do it in this PR but it's too code change for something that won't make it into core right now.

src/utils/filter_methods.ts
arielmelendez: YES, and now the question is what tests aren't failing.

arielmelendez:
> Getting two "hidden" errors during the `yarn test` execution:
> 
> ```shell
> Error: Error from inside the action
> error: required option '-p --required-type-parameter <requiredTypeParameter>' not specified
> ```
> 
> Not critical, but I'm a bit perplexed why we still have errors hidden from mocha

The latter is "expected" from the ParametersHelper test, though not ideal. The former looks new. I'll check it out.

src/utils/filter_methods.ts
arielmelendez: The answer is that there are no tests since this (and perhaps others of the) filter(s) function(s) are used in read-only workflows that aren't setup for testability yet 😑 I'll add some and patch this.

arielmelendez: Actually it works because the filter functions are used in a way where you'll get referential equality on everything, but we should still test and update them.

arielmelendez: Updated in 1698875 thru a4a7dde

arielmelendez:
> > Getting two "hidden" errors during the `yarn test` execution:
> > ```shell
> > Error: Error from inside the action
> > error: required option '-p --required-type-parameter <requiredTypeParameter>' not specified
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not critical, but I'm a bit perplexed why we still have errors hidden from mocha
> 
> The latter is "expected" from the ParametersHelper test, though not ideal. The former looks new. I'll check it out.

```
Error: Error from inside the action

  The CLIAction class
    ✔ throws an error when runningAction is called before the `trigger` method
```
I believe this is expected.

fedellen (dismissed):

matibat (approved):
It's ready to go!

fedellen:
We've got conflicts here now with upsert/skip heading for release, consider merging this integration branch where I merged these branches: https://github.com/ardriveapp/ardrive-cli/pull/151  

It should resolve the conflicts

fedellen (dismissed):

fedellen (approved):

fedellen (approved):

PR #141, opened 2021-11-05
PE-661: send signed tx

arielmelendez:
Quick hack together of a neat user-submitted idea.


fedellen (approved):
Steps in the README worked for me!

This seems simple enough to live on the CLI side here.

package.json
```diff
@@ -12,6 +12,7 @@
 		"arweave": "^1.10.16",
 		"commander": "^8.2.0",
 		"lodash": "^4.17.21",
+		"node-fetch": "^2.6.1",
```
arielmelendez, line 15: @fedellen are we going with axios from now on? Should I update the PR to use it instead?

arielmelendez: 0fb45b1

fedellen: `node-fetch` was excised from Core side in favor of `axios` for better browser compatibility. We are free to use whatever fetching methods we want here on CLI side.

However, sticking with axios here will reduce dependency size since its already included with Core and Arweave 👍

fedellen (dismissed):

fedellen (approved):

PR #142, opened 2021-11-05
Updated ArFS Link

jordanschau:
Link was broken and I believe the new link is correct.


fedellen (approved):
Thanks!

matibat (approved):
Good catch! Thank you @jordanschau

fedellen:
Merging this up to edit the link, we will use a link we think will be more permanent (https://ardrive.atlassian.net/l/c/m6P1vJDo) 

Thank you again for the PR and finding the problem!

PR #143, opened 2021-11-08
PE-635 Bug Fixes

arielmelendez:
•Needed to make sure destructured options variables got fed to type-safe constructors before use with ArDrive class.
• Needed to replace all usages of .includes on arrays of objects to use .some instead.

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -378,7 +378,7 @@ export class ArDrive extends ArDriveAnonymous {
 			owner
 		});
 
-		if (childrenFolderIds.includes(newParentFolderId)) {
+		if (childrenFolderIds.some((fid) => fid.equals(newParentFolderId))) {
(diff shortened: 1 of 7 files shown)


src/ardrive.ts
```diff
@@ -378,7 +378,7 @@ export class ArDrive extends ArDriveAnonymous {
 			owner
 		});
 
-		if (childrenFolderIds.includes(newParentFolderId)) {
+		if (childrenFolderIds.some((fid) => fid.equals(newParentFolderId))) {
```
matibat, line 381: I didn't know this method. I assume it's similar to find, but doesn't ensure to return the 1st result

matibat (approved):
It seems good. I still have to test it

src/ardrive.ts
arielmelendez: It's as efficient but returns a boolean rather than an entry

fedellen (approved):

PR #146, opened 2021-11-09
PE-672: Use object params on ArDrive class methods for extensibility

fedellen:
This PR refactors all of the public methods in the ArDrive and ArDrive anonymous classes to use the object de-structuring pattern on parameters. This will allow us to further extend each of these methods without resorting to long and potentially error-prone positional params.

The destination branch is a newly created integration branch between the safe types branch and the upload conflict branch to avoid conflicts when they merge

### DIFF
--- a/src/ardrive.ts
+++ b/src/ardrive.ts
@@ -2,7 +2,19 @@ import { ArFSDAO } from './arfsdao';
 import { UploadPrivateFileParams, UploadPublicFileParams } from './ardrive.types';
 import { CommunityOracle } from './community/community_oracle';
 import { deriveDriveKey, DrivePrivacy, GQLTagInterface } from 'ardrive-core-js';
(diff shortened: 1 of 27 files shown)


src/ardrive.ts
```diff
@@ -1536,11 +1460,11 @@ export class ArDrive extends ArDriveAnonymous {
 		};
 	}
 
-	async getDriveIdForFileId(fileId: FileID): Promise<DriveID> {
+	public async getDriveIdForFileId({ fileId }: GetDriveIdForFileIdParams): Promise<DriveID> {
```
arielmelendez, line 1463: Seems like overkill here.

fedellen: b26cf47

arielmelendez (changes requested):
Test

arielmelendez (approved):

PR #147, opened 2021-11-10
PE-463: replaces the === with .equals() for entity id comparison

matibat:


fedellen (approved):
Nice catch!

matibat:
This PR adds no change to the logic. We are fine just as it is. CLosing...

PR #150, opened 2021-11-10
PE-675: Use consistent tags on tip transactions

### LINKED ISSUE #136: Inconsistent tag usage for tip/fee transactions
The new CLI creates tip transactions with a tag `Type:data upload` instead of following the established convention of having two tags, `Type:fee` and `Tip-Type:data upload`.

fedellen:
This PR changes the tip transaction tags to be consistent with the web app.

Fixes #136

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "ardrive-cli",
-	"version": "1.0.2",
+	"version": "1.0.3",
 	"description": "The ArDrive Command Line Interface (CLI is a Node.js application for terminal-based ArDrive workflows. It also offers utility operations for securely interacting with Arweave wallets and inspecting various Arweave blockchain conditions.",
 	"main": "./lib/index.js",
 	"bin": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "ardrive-cli",
-	"version": "1.0.2",
+	"version": "1.0.3",
```
arielmelendez, line 3: Nice! Though we should have done this immediately after the last release.

src/commands/send_ar.ts
```diff
@@ -34,8 +34,8 @@ new CLICommand({
 			rewardSetting,
 			options.dryRun,
 			[
-				{ name: 'App-Name', value: 'ArDrive-CLI' },
-				{ name: 'App-Version', value: '2.0' },
+				{ name: 'App-Name', value: CLI_APP_NAME },
```
arielmelendez, line 37: Nice

arielmelendez (approved):

package.json
fedellen: Should we do a version PR to dev after each release?

PR #151, opened 2021-11-10
PE-638 + PE-635 Integration

fedellen:
This PR merges together the skip/upsert branch and the safe types integration branch. With skip/upsert now on `dev`, this should resolve the conflicts now present in `PE-635_integration`


src/parameter_declarations.ts
```diff
@@ -243,6 +249,31 @@ Parameter.declare({
...
+	description: '(OPTIONAL) Create new file revisions if there is a name conflict within destination folder',
+	type: 'boolean',
+	forbiddenConjunctionParameters: [SkipParameter, UpsertParameter]
+});
+
+Parameter.declare({
+	name: UpsertParameter,
+	aliases: ['--upsert'],
+	description:
+		'(OPTIONAL) When there is a name conflict within the destination folder, if that file was last modified at the same time as the file to upload, skip the upload, otherwise upload that file as a new revision',
```
arielmelendez, line 272: Can we generalize the language here to consider that this could be applied to file or folder uploads?

arielmelendez (approved):

src/parameter_declarations.ts
fedellen: Actually, upsert logic doesn't really effect folder conditions. As per the [table](https://github.com/ardriveapp/ardrive-cli/tree/dev#name-conflict-resolution-on-upload), folder to X conflicts shares the behaviors of replace. And the only difference it with skip is that it will `Fail` instead of skipping.

This message should be simplified though, how about:

```suggestion
		"(OPTIONAL) Update files only when they've been modified if there is a name conflict within the destination folder",
```

arielmelendez: The language is a little clunky. Reorder to:
```
(OPTIONAL)  When there is a name conflict within the destination folder, only upload file if a modification is detected. Skip otherwise.
```

fedellen: 9fff5cc

PR #155, opened 2021-11-11
fix(create folder): Assert drive privacy during folder creation PE-679

fedellen:
This PR asserts the drive privacy of a folder creation during the GQL call to derive the drive owner.

As this is a hotfix, the version has been bumped here and the target branch is `master`. Proposed GitHub release notes:

- Creating private entities inside public drives and public entities inside private drives will now cause an error
- Creating private entities with the wrong drive key or password will now cause an error

### DIFF
--- a/src/arfsdao_anonymous.ts
+++ b/src/arfsdao_anonymous.ts
@@ -46,7 +46,13 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 	}
 
 	public async getOwnerForDriveId(driveId: DriveID): Promise<ArweaveAddress> {
-		const gqlQuery = buildQuery({ tags: [{ name: 'Drive-Id', value: `${driveId}` }], sort: ASCENDING_ORDER });
+		const gqlQuery = buildQuery({
+			tags: [
+				{ name: 'Drive-Id', value: `${driveId}` },
+				{ name: 'Entity-Type', value: 'drive' }
+			],
+			sort: ASCENDING_ORDER
+		});
 		const response = await this.arweave.api.post(graphQLURL, gqlQuery);
 		const edges: GQLEdgeInterface[] = response.data.data.transactions.edges;
 
--- a/src/arfsdao_anonymous.ts
+++ b/src/arfsdao_anonymous.ts
@@ -87,7 +87,9 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 					throw new Error('Target private drive has no "Cipher-IV" tag!');
 				}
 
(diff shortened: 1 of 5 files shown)


src/arfsdao_anonymous.ts
```diff
@@ -55,6 +63,14 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
 		}
 
 		const edgeOfFirstDrive = edges[0];
+
+		if (drivePrivacy) {
+			// Assert drive privacy tag if it has been passed down
+			if (edgeOfFirstDrive.node.tags.find((t) => t.name === 'Drive-Privacy')?.value !== drivePrivacy) {
```
arielmelendez, line 69: It would be problematic if there were no Drive-Privacy tag. Perhaps we should error on that?

arielmelendez (dismissed):

src/arfsdao_anonymous.ts
fedellen: In that case it will error here, `undefined` will not equal `drivePrivacy`

fedellen: But we could give a more specific error, adding one

fedellen: 23e06bc

arielmelendez (dismissed):

src/arfsdao_anonymous.ts
```diff
@@ -46,14 +46,18 @@ export class ArFSDAOAnonymous extends ArFSDAOType {
...
-		return this.getOwnerForDriveId(driveId, 'private');
+	public async getOwnerForPrivateDriveId(driveId: DriveID, driveKey: DriveKey): Promise<ArweaveAddress> {
+		return this.getOwnerForDriveId(driveId, true, driveKey);
 	}
 
-	public async getOwnerForDriveId(driveId: DriveID, drivePrivacy?: DrivePrivacy): Promise<ArweaveAddress> {
+	public async getOwnerForDriveId(
+		driveId: DriveID,
+		assertPrivacy = false,
+		driveKey?: DriveKey
```
arielmelendez, line 59: So unfortunately this violates the "anonymous" part of of this class. We need to hoist the private versions of this function into the ArFSDAO layer.

fedellen: I've addressed this here: b81960c

Let me know what you think!

arielmelendez (dismissed):

agsuy (approved):

arielmelendez (approved):

PR #158, opened 2021-11-16
PE-699: Release ArDrive CLI v1.1.0

fedellen:
Proposed GitHub release changelog:

- All core architecture has been refactored into ardrive-core-js
- Several dependencies have been excised
- User provided values (e.g. drive id, arweave address) will now be safely checked and provide detailed errors


matibat (approved):
Ready to ship! :clap:

src/index.ts
```diff
@@ -1,15 +1,14 @@
-#!/usr/bin/env node
```
fedellen, line 1: Oh no

PR #162, opened 2021-11-18
refactor(stubs): Use CLI side stub PE-701

fedellen:
This PR removes the stub imported by Core, because it is no longer being exported. Our tests should not rely on stubs from upstream dependencies anyhow


fedellen:
This build is expected to fail until Core 1.0.1 exists

agsuy (approved):
LGTM!

PR #168, opened 2021-11-29
PE-464: download-file 

matibat:


src/commands/download_file.ts
```diff
@@ -0,0 +1,36 @@
...
+new CLICommand({
+	name: 'download-file',
+	parameters: [FileIdParameter, LocalFilePathParameter, DriveIdParameter, ...DrivePrivacyParameters],
+	action: new CLIAction(async (options) => {
+		const parameters = new ParametersHelper(options);
+		const fileId = parameters.getRequiredParameterValue(FileIdParameter, EID);
+		const destFolderPath = resolvePath(parameters.getParameterValue(LocalFilePathParameter) || './');
+
+		if (await parameters.getIsPrivate()) {
+			const driveId = parameters.getRequiredParameterValue(DriveIdParameter);
```
matibat, line 22: What if I provide the drive key as parameter? Drive ID shouldn't be required at all

fedellen: We should probably just eat the `getDriveIdForFileId` GQL call rather than having a driveID param. That is the pattern that most of the other commands are using.

You do bring up a great point here in that case too though, we wouldn't even need the drive ID if they supplied a driveKey. That inefficiency shows up in other commands where we don't pass down a driveId too, like `file-info` for example.

fedellen:
We should have a README component here. At the very least we can add it to the command list at the bottom and maybe add a download section later.

fedellen:
I ran a quick dev test with a local Core @ `PE-464_download_file`. The file seems to have successfully downloaded but I also ran into this error here:

```shell
 yarn ardrive download-file -f ff450770-a9cb-46a5-9234-89cbd9796610 -d bc9af866-6421-40f1-ac89-202bddb5c487                      File with ID "ff450770-a9cb-46a5-9234-89cbd9796610" was successfully download to "/Users/derek/ardrive/ardrive-cli"
(node:8109) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'fileEntity' of undefined
    at setLastModifiedDate (/Users/derek/ardrive/ardrive-cli/.yarn/unplugged/ardrive-core-js-file-4af702b991/node_modules/ardrive-core-js/lib/arfs/arfs_file_wrapper.js:228:60)
    at <anonymous>
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:8109) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:8109) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

package.json
```diff
@@ -8,7 +8,7 @@
 	},
 	"types": "./lib/index.d.ts",
 	"dependencies": {
-		"ardrive-core-js": "1.0.4",
+		"ardrive-core-js": "../ardrive-core-js_PE-464_download_file",
```
arielmelendez, line 11: TODO: Update

src/commands/download_file.ts
arielmelendez: I agree. Let's make driveKey and driveID optional for this command if possible.

src/commands/download_file.ts
```diff
@@ -0,0 +1,36 @@
...
+			const driveKey = await parameters.getDriveKey({ driveId: EID(driveId) });
+			const wallet = await parameters.getRequiredWallet();
+			const ardrive = cliArDriveFactory({
+				wallet,
+				feeMultiple: parameters.getOptionalBoostSetting()
+			});
+			await ardrive.downloadPrivateFile(fileId, destFolderPath, driveKey);
```
arielmelendez, line 29: What do you guys think of this behavior for -l:
• if the path is an existing folder path on disk, deposit the file in that folder based on its current name on ArDrive
• if the path is NOT an existing folder path on disk, assume the user wants to use it as the local file path, ignoring the file's name on ArDrive. To Discuss: Should we create any needed folder paths on the way to the file if they don't exist?

arielmelendez (changes requested):
See my comments

package.json
matibat: That core branch is the right one

src/commands/download_file.ts
matibat: This was already applied :+1:

src/commands/download_file.ts
```diff
@@ -19,13 +19,14 @@ new CLICommand({
...
 
 		if (await parameters.getIsPrivate()) {
-			const driveId = parameters.getRequiredParameterValue(DriveIdParameter);
-			const driveKey = await parameters.getDriveKey({ driveId: EID(driveId) });
 			const wallet = await parameters.getRequiredWallet();
 			const ardrive = cliArDriveFactory({
 				wallet,
 				feeMultiple: parameters.getOptionalBoostSetting()
 			});
+			const driveIdString = parameters.getParameterValue(DriveIdParameter);
```
fedellen, line 27: This is still required at the parameter declaration level

fedellen: Even for public downloads

fedellen:
Downloading a 200MB private file I ran into this error:

```shell
Error: Premature close
```

But it only happened on the first run, haven't been able to reproduce it to print the stack trace. It worked on the next 5 attempts

Might've been an environment issue, but leaving this here in case you have seen this before

src/commands/download_file.ts
fedellen: That behavior might feel more natural to a user. 

Right now it seems to be only used as the folder path. If we keep this current behavior for now, I think we should at least use a new parameter with the name DestinationFolderPath or similar so that users don't confuse it with a file name parameter.

>  To Discuss: Should we create any needed folder paths on the way to the file if they don't exist?

I think yes. Creating folders for the user provides convenience and should be arbitrary with `fs` or similar

src/commands/download_file.ts
```diff
@@ -0,0 +1,37 @@
...
+import {
+	DriveIdParameter,
+	DrivePrivacyParameters,
+	FileIdParameter,
+	LocalFilePathParameter
+} from '../parameter_declarations';
+
+new CLICommand({
+	name: 'download-file',
+	parameters: [FileIdParameter, LocalFilePathParameter, DriveIdParameter, ...DrivePrivacyParameters],
```
fedellen, line 15: Make this non-required here or remove the DriveID param altogether

```suggestion
	parameters: [
		FileIdParameter,
		LocalFilePathParameter,
		{ name: DriveIdParameter, required: false },
		...DrivePrivacyParameters
	],
```

fedellen: And probably switch `LocalFilePathParameter` for a new `DestFolderPathParameter` since that is the behavior its currently using

fedellen (changes requested):
A couple of small things. DriveID still seems to be required and the help output for the `-l` param here could be confusing to a user:

```shell
  -l --local-file-path <localFilePath>              the path on the local filesystem for the file that will be uploaded
```

src/commands/download_file.ts
matibat: > To Discuss: Should we create any needed folder paths on the way to the file if they don't exist?

I'd rather add a flag to force creating the directory tree the same way `mkdir -p` does. If the parent folder does not exist and no recursive flag provided, then throw the default `ENOENT`

src/commands/download_file.ts
matibat: 95e614a

matibat: d86bb4f7a9784e34c2bd40092b72c0e9cc07886b

matibat: Removed

src/commands/download_file.ts
matibat: See the commit https://github.com/ardriveapp/ardrive-core-js/pull/91/commits/2412b211c28d722c3b99d392d644631773cdd872 at core for the localFilepathResolution algorithm

README.md
```diff
@@ -1005,6 +1005,7 @@ list-folder
 list-drive
 list-all-drives
 
+download-file
```
matibat, line 1008: TODO: add doc

fedellen (dismissed):
👍

package.json
arielmelendez: Thanks, but the reminder is just to update this back to the correct npm release before merging :)

src/commands/download_file.ts
arielmelendez: > I'd rather add a flag to force creating the directory tree the same way mkdir -p does. If the parent folder does not exist and no recursive flag provided, then throw the default ENOENT

Let's make a decision based on the ease of use of the following permutations:
1) -l path is a folder on the local file system that exists
    • Result: CLI uses folder as destPath and provides an undefined filename, leading to the ArDrive filename being used in the full path.
2) -l path is a file on the local file system that exists
    • Result: CLI uses the parent folder of the file as destPath and provides the filename. ArDrive filename is ignored.
3) -l path is a nonexistent entity on the local file system, but whose parent folder exists
    • Result: CLI uses the parent folder of the entity as destPath and provides the entity name as a filename. ArDrive filename is ignored. (DOES THIS SEEM CORRECT?!?)
3) -l path is a nonexistent entity on the local file system whose parent folder does NOT exist
    • Result: ERROR

Let's see how a flag like `mkdir`'s `-p` would change these cases:
1) No change
2) No change
3) No change, DEPENDING on whether we think the behavior described above is desirable
4) The path to the parent folder of the entity is created. The entity name is used. ArDrive name is ignored.

@agsuy should weigh in here.

src/commands/download_file.ts
```diff
@@ -0,0 +1,33 @@
...
+
+new CLICommand({
+	name: 'download-file',
+	parameters: [FileIdParameter, DestinationOutputPathParameter, ...DrivePrivacyParameters],
+	action: new CLIAction(async (options) => {
+		const parameters = new ParametersHelper(options);
+		const fileId = parameters.getRequiredParameterValue(FileIdParameter, EID);
+		const destOutputPath = parameters.getParameterValue(DestinationOutputPathParameter) || './';
+
+		const [destFolderPath, fileName] = getOutputFilePathAndName(destOutputPath);
```
matibat, line 16: Create a type for the returned arr

src/parameter_declarations.ts
```diff
@@ -231,6 +232,13 @@ Parameter.declare({
 	description: `the path on the local filesystem for the file that will be uploaded`
 });
 
+Parameter.declare({
+	name: DestinationOutputPathParameter,
+	aliases: ['-d', '--dest-output-path'],
+	description: `the path on the local filesystem where to download`,
```
matibat, line 238: Rephrase this

src/utils/local_file_path.ts
```diff
@@ -0,0 +1,25 @@
+import { statSync } from 'fs';
+import { basename, dirname, resolve } from 'path';
+
+export function getOutputFilePathAndName(destOutputPath: string): [string, string?] {
```
matibat, line 4: Add docblock & set the return type

src/commands/download_file.ts
arielmelendez: Ideas for the flag, if we ever use it:
--create-paths
--parents
--create-intermediates alias: -i

src/parameter_declarations.ts
arielmelendez:
```suggestion
	description: `a destination path for the file. Uses same destination rules as cp, i.e.:
\t\t\t\t\t\t\t-l /path/to/existing/folder (on-chain filename used in existing folder)
\t\t\t\t\t\t\t-l /path/to/existing/file (on-chain filename ignored, local file overwritten)
\t\t\t\t\t\t\t-l /path/to/existing/folder/nonexistent_thing (nonexistent_thing used as filename)
\t\t\t\t\t\t\t-l /path/to/nonexistent/folder (error)
`,
```

src/utils/local_file_path.ts
matibat: 223cd4e328a496788dff91737832bd612ab359ee

src/commands/download_file.ts
matibat: 223cd4e328a496788dff91737832bd612ab359ee

README.md
matibat: 19a4cee745be160ec1525fe08aac66c4e003e23f

matibat:
Re-updated the readme. Thanks for your suggestions @arielmelendez

fedellen (approved):
Working well on my end!

arielmelendez (approved):

fedellen (approved):

src/parameter_declarations.ts
```diff
@@ -20,6 +20,7 @@ export const FolderIdParameter = 'folderId';
 export const FileIdParameter = 'fileId';
 export const ParentFolderIdParameter = 'parentFolderId';
 export const LocalFilePathParameter = 'localFilePath';
+export const DestinationOutputPathParameter = 'destOutputPath'; // TODO: find a better name
```
matibat, line 23: `--local-path` !

src/parameter_declarations.ts
```diff
@@ -233,14 +233,13 @@ Parameter.declare({
...
-	description: `a destination path for the file. Uses same destination rules as cp, i.e.:
-\t\t\t\t\t\t\t-d /path/to/existing/folder (on-chain filename used in existing folder)
-\t\t\t\t\t\t\t-d /path/to/existing/file (on-chain filename ignored, local file overwritten)
-\t\t\t\t\t\t\t-d /path/to/existing/folder/nonexistent_thing (nonexistent_thing used as filename)
-\t\t\t\t\t\t\t-d /path/to/nonexistent/folder (error)
-`,
+	name: LocalFilePathDownloadParameter,
+	aliases: ['--local-path'],
+	description: `the path on the local filesystem for the file to be download
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-file-path
```
arielmelendez, line 239: I don't think anything from here down matters.

matibat: f9171f2

arielmelendez (approved):

arielmelendez (dismissed):

arielmelendez (approved):

src/commands/download_file.ts
```diff
@@ -0,0 +1,43 @@
...
+import { getOutputFilePathAndName } from '../utils';
+import { join } from 'path';
+
+new CLICommand({
+	name: 'download-file',
+	parameters: [FileIdParameter, LocalFilePathDownloadParameter, ...DrivePrivacyParameters],
+	action: new CLIAction(async (options) => {
+		const parameters = new ParametersHelper(options);
+		const fileId = parameters.getRequiredParameterValue(FileIdParameter, EID);
+		const destOutputPath = parameters.getParameterValue(LocalFilePathDownloadParameter) || './';
```
matibat, line 15: Windows won't be happy with that extra `/`
```suggestion
		const destOutputPath = parameters.getParameterValue(LocalFilePathDownloadParameter) || '.';
```

arielmelendez (approved):

fedellen (approved):

arielmelendez (approved):

arielmelendez (dismissed):

fedellen (approved):

PR #169, opened 2021-11-29
PE-739: Implement an option to use glob on upload-file command

karlprieb:
* Add `-g`/`--glob` parameter to upload-file command
* Standardize  output results for when you upload multiple files at once without uploading a folder.

Result output example:
```json
{
    "created": [
        {
            "type": "file",
            "metadataTxId": "mhyOGksQHxY3puiyq639DOJBPaUgPrOPlwBGoKyxsn4",
            "dataTxId": "-TrMMOeYlNgzguSjtEZOTxMsW104Q8pso6fDHTo7Lkw",
            "entityId": "61f4419f-b00d-40e0-ab22-e9e3a83092f3"
        },
        {
            "type": "file",
            "metadataTxId": "r0dEbDIYg4TVeckDUQOq0vo9jBV7iTQ0iCaEXU3v99w",
            "dataTxId": "Rem2gjwmwfi3PmHP_OqR8fRB2tVoFIalZIszJT4O1VI",
            "entityId": "ea187b39-6920-45eb-95f8-0680179410ca"
        }
    ],
    "tips": [
        {
            "txId": "FFsYt8ymzc9Dk0-QFNjtPUPIFEIgS0vOpFHPa-ehHms",
            "recipient": "Zznp65qgTIm2QBMjjoEaHKOmQrpTu0tfOcdbkm_qoL4",
            "winston": "10000000"
        },
        {
            "txId": "UPWj53A2iEXtrSCwveKjqCExyWiRkDKtBGVNgNpLilk",
            "recipient": "Zznp65qgTIm2QBMjjoEaHKOmQrpTu0tfOcdbkm_qoL4",
            "winston": "10000000"
        }
    ],
    "fees": {
        "-TrMMOeYlNgzguSjtEZOTxMsW104Q8pso6fDHTo7Lkw": "43942624",
        "mhyOGksQHxY3puiyq639DOJBPaUgPrOPlwBGoKyxsn4": "43942624",
        "FFsYt8ymzc9Dk0-QFNjtPUPIFEIgS0vOpFHPa-ehHms": "531578",
        "Rem2gjwmwfi3PmHP_OqR8fRB2tVoFIalZIszJT4O1VI": "43942624",
        "r0dEbDIYg4TVeckDUQOq0vo9jBV7iTQ0iCaEXU3v99w": "43942624",
        "UPWj53A2iEXtrSCwveKjqCExyWiRkDKtBGVNgNpLilk": "531578"
    }
}
```


README.md
```diff
@@ -703,6 +703,15 @@ Users can perform a bulk upload by using the upload-file command on a target fol
 ardrive upload-file --local-file-path /path/to/folder  --parent-folder-id "9af694f6-4cfc-4eee-88a8-1b02704760c0" -w /path/to/wallet.json
 ```
 
+### Uploading files using a glob pattern (BETA)
+
+Users can declare a glob pattern to select files to be upload in bulk. `upload-file` command will parse the glob pattern and will upload all the files that match the pattern to the selected destination.
+The example above will upload all `jpg` files from the indicated path.
+
+```shell
+ardrive upload-file --glob "/path/to/photos/folder/*.jpg" --parend-folder-id "9af694f6-4cfc-4eee-88a8-1b02704760c0" -w /path/to/wallet.json
```
fedellen, line 712: spelling nit:

```suggestion
ardrive upload-file --glob "/path/to/photos/folder/*.jpg" --parent-folder-id "9af694f6-4cfc-4eee-88a8-1b02704760c0" -w /path/to/wallet.json
```

fedellen:
Very cool stuff!

I must note here that this will suffer from two of the same inefficiencies that our current iteration of CSV upload. The first, it will send a community tip for each file uploaded rather than sending one tip for an entire bulk upload. And the other, it will send unnecessary GQL calls to retrieve drive information for every file. These are not things that can be fixed on the CLI side alone, but rather something we will need to account for in Core with an `uploadAllTheseFiles` method.

I don't think that holds this back though, I'm fine with `--glob` uploads being on par with CSV uploads for now.

arielmelendez:
> Very cool stuff!
> 
> I must note here that this will suffer from two of the same inefficiencies that our current iteration of CSV upload. The first, it will send a community tip for each file uploaded rather than sending one tip for an entire bulk upload. And the other, it will send unnecessary GQL calls to retrieve drive information for every file. These are not things that can be fixed on the CLI side alone, but rather something we will need to account for in Core with an `uploadAllTheseFiles` method.
> 
> I don't think that holds this back though, I'm fine with `--glob` uploads being on par with CSV uploads for now.

Great points @fedellen . I think we should release this as a feature in beta and make the changes you suggested in core. It's probably time to add that ArFS in memory cache there as well.

karlprieb:
> Very cool stuff!
> 
> I must note here that this will suffer from two of the same inefficiencies that our current iteration of CSV upload. The first, it will send a community tip for each file uploaded rather than sending one tip for an entire bulk upload. And the other, it will send unnecessary GQL calls to retrieve drive information for every file. These are not things that can be fixed on the CLI side alone, but rather something we will need to account for in Core with an `uploadAllTheseFiles` method.
> 
> I don't think that holds this back though, I'm fine with `--glob` uploads being on par with CSV uploads for now.

You're right. Its not efficient, but as @arielmelendez mentioned we will need a new method for this type of upload. I can take a look on core if you're OK with that. It will take some time as I'm learning on my spare time.

arielmelendez:
>  I can take a look on core if you're OK with that. 
@karlprieb , I'd hold off for now. We're about to jump into bundling with ANS-104 bundles (direct to network, NOT through Bundlr) and that will require us to carefully craft the multi-file interfaces for ardrive-core. Please feel free to experiment locally, though, especially if it'll reduce upload costs for you in the near term!

karlprieb:
> > I can take a look on core if you're OK with that.
> 
> @karlprieb , I'd hold off for now. We're about to jump into bundling with ANS-104 bundles (direct to network, NOT through Bundlr) and that will require us to carefully craft the multi-file interfaces for ardrive-core. Please feel free to experiment locally, though, especially if it'll reduce upload costs for you in the near term!

OK :) I didn't km that's coming on a near future! Will be awesome!

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 14.17.0
```
arielmelendez, line 1: Is this for asdf? If so, it seems the nodejs plugin already supports reading from .nvmrc which we have in this repo. See: https://github.com/asdf-vm/asdf-nodejs/
In which case, I don't think this file would be necessary. If you agree, could you please remove it from this PR? We'd just like to avoid duplicating the version numbers we need in multiple places. Thanks!

README.md
```diff
@@ -703,6 +703,15 @@ Users can perform a bulk upload by using the upload-file command on a target fol
 ardrive upload-file --local-file-path /path/to/folder  --parent-folder-id "9af694f6-4cfc-4eee-88a8-1b02704760c0" -w /path/to/wallet.json
 ```
 
+### Uploading files using a glob pattern (BETA)
```
arielmelendez, line 706: Thanks for adding this section! You'll also need to add a link to it in the table of contents above. Since the heading is rather long, use an anchor with an id so that you can easily link to it from the ToC, e.g. like the above:
```
### Uploading a Folder with Files (Bulk Upload)<a id="bulk-upload"></a>
```

README.md
```diff
@@ -703,6 +703,15 @@ Users can perform a bulk upload by using the upload-file command on a target fol
...
+
+Users can declare a glob pattern to select files to be upload in bulk. `upload-file` command will parse the glob pattern and will upload all the files that match the pattern to the selected destination.
+The example above will upload all `jpg` files from the indicated path.
```
arielmelendez, line 709: `The example above` --> `The example below`

package.json
```diff
@@ -18,6 +18,7 @@
 	"devDependencies": {
 		"@istanbuljs/nyc-config-typescript": "^1.0.1",
 		"@types/chai": "^4.2.21",
+		"@types/glob": "^7",
```
arielmelendez, line 21: Any concerns with letting the types version here float apart from the version number?

src/commands/upload_file.ts
```diff
@@ -32,20 +35,40 @@ interface UploadFileParameter {
...
 }
 
+type FilePath = string;
+
+const formatResults = (results: ArFSResult[]): ArFSResult => {
+	const isFolderUploadResults = results.length === 1;
+	if (isFolderUploadResults) {
+		return results[0];
+	}
+	return results.reduce(
```
arielmelendez, line 45: Doesn't think generalize to the results.length === 1 case? In which case the special case above goes away.

src/commands/upload_file.ts
```diff
@@ -32,20 +35,40 @@ interface UploadFileParameter {
...
+		GlobParameter,
+		LocalFilePathParameter,
+		LocalFilesParameter,
+		ParentFolderIdParameter,
 		...ConflictResolutionParams,
 		...DrivePrivacyParameters
 	],
 	action: new CLIAction(async function action(options) {
 		const parameters = new ParametersHelper(options);
+		const globValue = parameters.getParameterValue(GlobParameter);
```
arielmelendez, line 70: Nit: could improve the locality of reference by moving this to where it's used down below.

src/parameter_declarations.ts
```diff
@@ -32,14 +32,15 @@ export const ReplaceParameter = 'replace';
...
+export const GlobParameter = 'glob';
 
 // Aggregates for convenience
 export const WalletTypeParameters = [WalletFileParameter, SeedPhraseParameter];
 export const DriveCreationPrivacyParameters = [...WalletTypeParameters, PrivateParameter, UnsafeDrivePasswordParameter];
 export const DrivePrivacyParameters = [DriveKeyParameter, ...DriveCreationPrivacyParameters];
 export const TreeDepthParams = [AllParameter, MaxDepthParameter];
 export const AllParameters = [
-	AddressParameter,
+  AddressParameter,
```
arielmelendez, line 43: Whitespace issue

src/parameter_declarations.ts
```diff
@@ -54,6 +55,7 @@ export const AllParameters = [
 	FolderIdParameter,
 	FolderNameParameter,
 	GetAllRevisionsParameter,
+  GlobParameter,
```
arielmelendez, line 58: Whitespace issue

arielmelendez (changes requested):
Requested a few more minor changes. Looking good! Grateful for the improvement to the output of the ArFSResult array!

.tool-versions
karlprieb: I didn't know that... make sense to remove

package.json
karlprieb: I don't think so, we can have a fixed minor version if you want, but if both are using the same major version it won't be a problem.
What do you think @arielmelendez ?

src/commands/upload_file.ts
karlprieb: We could have just the `reduce` as it works for both cases, but there's no need to iterate over results when you have only one result. I thought that it would be less expensive to just check for it instead of running the spread operators for nothing.

Do you think that makes the code more unreadable?

src/commands/upload_file.ts
karlprieb: Great pick... I let it there because at first I was using this const to change the results report.

README.md
```diff
@@ -703,6 +704,15 @@ Users can perform a bulk upload by using the upload-file command on a target fol
 ardrive upload-file --local-file-path /path/to/folder  --parent-folder-id "9af694f6-4cfc-4eee-88a8-1b02704760c0" -w /path/to/wallet.json
 ```
 
+### Uploading files using a glob pattern (BETA)<a id="glob-upload"></a>
+
+Users can declare a glob pattern to select files to be upload in bulk. `upload-file` command will parse the glob pattern and will upload all the files that match the pattern to the selected destination.
```
arielmelendez, line 709:
```suggestion
Users can declare a glob pattern to select files to be uploaded in bulk. The `upload-file` command will parse the glob pattern and will upload all the files that match the pattern to a specified parent folder.
```

src/commands/upload_file.ts
arielmelendez: I agree that your solution is more performant in the single item case! But I think it's trading off a very tiny bit of performance for more lines of code that need to be tested, read, and comprehended. I think the reduce is clear and performant "enough" :)

src/commands/upload_file.ts
```diff
@@ -32,20 +35,39 @@ interface UploadFileParameter {
 	driveKey?: DriveKey;
 }
 
+type FilePath = string;
+
+const formatResults = (results: ArFSResult[]): ArFSResult => {
```
arielmelendez, line 40: Not critical, but a set of unit tests for this could be nice. I'd approve without it though.

fedellen (dismissed):

arielmelendez (dismissed):
Thanks @karlprieb ! We'll test internally as soon as we can :)

src/commands/upload_file.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import { expect } from 'chai';
+import { ArFSResult, ArweaveAddress, EntityID, TransactionID, Winston } from 'ardrive-core-js';
+
+import { formatResults } from './upload_file';
+
+const createResultMock = (): ArFSResult => ({
+	created: [
+		{
+			type: 'file',
+			metadataTxId: (randomUUID() as unknown) as TransactionID,
```
karlprieb, line 11: @fedellen @arielmelendez I was not sure if it's ok to do that or if it's expected that I create a real TransactionID. As it's a test and the results doesn't matter (they just need to be different) I went with the easy way :D

src/commands/upload_file.test.ts
```diff
@@ -0,0 +1,54 @@
...
+		[randomUUID()]: new Winston(40121526),
+		[randomUUID()]: new Winston(40121526),
+		[randomUUID()]: new Winston(401226)
+	}
+});
+
+describe('Commands - upload-file', () => {
+	it('formatResults should merge multiple results in just one object', () => {
+		const firstResult = createResultMock();
+		console.log(firstResult.created[0].metadataTxId);
```
fedellen, line 33: Console log straggler

karlprieb: Ooops

src/commands/upload_file.test.ts
fedellen: Should be fine for this as the values of these fields aren't relevant to the test. But it could confuse some dev into later thinking that transaction IDs are UUIDs, which they are not. Maybe add a comment specifying that these TxIDs and the ArweaveAddress are invalid, but their validity is not important to the test.

 Thanks for adding coverage for this method!

karlprieb: Sure :D I was thinking on using something like https://github.com/fent/randexp.js to generate strings from the regex that ardrive-core-js have, but I thought that it's not worth it :D 
I'll add those comments.

fedellen (dismissed):

package.json
arielmelendez: Should be fine as long as it's using semantic versioning adequately 🤞

src/commands/upload_file.test.ts
arielmelendez: So I have a few issues with this approach:
image

1) Prettier and ESLint seem to be fighting over how this should be formatted. One wants parens around the `unknown` cast, and the other does not, resulting in a Problem in the IDE.
2) The choice to do things this way disguised a small bug on line 45:
```
		const firstFeeFromSecondResult = Object.keys(firstResult.fees)[0];
```
should be:
```
		const firstFeeFromSecondResult = Object.keys(secondResult.fees)[0];
```
But the test passes either way, and for at least two reasons that I can think of:
a) since we're not using literals, it's hard to see where something could go wrong
b) We're writing programmatic code that changes on each test run and hoping that the code we write to handle the values generated by that code are both correct AND actually test what needs to be tested.

I think it would be much simpler and straightforward to use some POJO literals with easy-to-see-and-comprehend values to validate that the function works as expected.

For what it's worth, there's a single stub exported from ardrive-core-js. But I don't think it's flexible enough or worth the effort of making flexible over, say, just creating the objects in place, to use here where we need multiple distinct values to be sure we're getting back the right values.
```
export const stubTransactionID = TxID('0000000000000000000000000000000000000000000');
```

src/commands/upload_file.test.ts
```diff
@@ -0,0 +1,55 @@
+import { randomUUID } from 'crypto';
```
arielmelendez, line 1: A suggestion for simplification based on my comments below:
```suggestion
import { expect } from 'chai';
import { ADDR, ArFSResult, EID, TxID, W } from 'ardrive-core-js';

import { formatResults } from './upload_file';

describe('Commands - upload-file', () => {
	it('formatResults should merge multiple results in just one object', () => {
		const firstResult: ArFSResult = {
			created: [
				{
					type: 'file',
					metadataTxId: TxID('0000000000000000000000000000000000000000000'),
					dataTxId: TxID('0000000000000000000000000000000000000000001'),
					entityId: EID('00000000-0000-0000-0000-000000000000'),
					key: 'fakeKey1'
				}
			],
			tips: [
				{
					txId: TxID('0000000000000000000000000000000000000000001'),
					recipient: ADDR('abcdefghijklmnopqrxtuvwxyz123456789ABCDEFGH'),
					winston: W(12345)
				}
			],
			fees: {
				'0000000000000000000000000000000000000000001': W(24680)
			}
		};

		const secondResult: ArFSResult = {
			created: [
				{
					type: 'file',
					metadataTxId: TxID('0000000000000000000000000000000000000000002'),
					dataTxId: TxID('0000000000000000000000000000000000000000003'),
					entityId: EID('00000000-0000-0000-0000-000000000001'),
					key: 'fakeKey2'
				}
			],
			tips: [
				{
					txId: TxID('0000000000000000000000000000000000000000003'),
					recipient: ADDR('123456789ABCDEFGHabcdefghijklmnopqrxtuvwxyz'),
					winston: W(54321)
				}
			],
			fees: {
				'0000000000000000000000000000000000000000003': W(13579)
			}
		};

		const resultsOutput = formatResults([firstResult, secondResult]);

		// created
		expect(resultsOutput).to.deep.equal({
			created: [
				{
					type: 'file',
					metadataTxId: TxID('0000000000000000000000000000000000000000000'),
					dataTxId: TxID('0000000000000000000000000000000000000000001'),
					entityId: EID('00000000-0000-0000-0000-000000000000'),
					key: 'fakeKey1'
				},
				{
					type: 'file',
					metadataTxId: TxID('0000000000000000000000000000000000000000002'),
					dataTxId: TxID('0000000000000000000000000000000000000000003'),
					entityId: EID('00000000-0000-0000-0000-000000000001'),
					key: 'fakeKey2'
				}
			],
			tips: [
				{
					txId: TxID('0000000000000000000000000000000000000000001'),
					recipient: ADDR('abcdefghijklmnopqrxtuvwxyz123456789ABCDEFGH'),
					winston: W(12345)
				},
				{
					txId: TxID('0000000000000000000000000000000000000000003'),
					recipient: ADDR('123456789ABCDEFGHabcdefghijklmnopqrxtuvwxyz'),
					winston: W(54321)
				}
			],
			fees: {
				'0000000000000000000000000000000000000000001': W(24680),
				'0000000000000000000000000000000000000000003': W(13579)
			}
		});
	});
});
```

arielmelendez: That did not work the way I wanted it to in GH 😂 But feel free to apply locally instead of using the suggestion feature!

src/commands/upload_file.test.ts
arielmelendez: Take a look at the suggestion I posted as an alternative above. Thanks!

src/commands/upload_file.test.ts
karlprieb: Suggestion accepted 🎉

src/parameter_declarations.ts
```diff
@@ -325,3 +327,12 @@ Parameter.declare({
 		'(OPTIONAL) Derives a drive key for the given drive ID without verifying its correctness against the drive on chain.',
 	type: 'boolean'
 });
+
+Parameter.declare({
+	name: GlobParameter,
+	aliases: ['-g', '--glob'],
+	description: `(BETA) a glob pattern to select files to be uploaded
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files-path
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files`,
```
arielmelendez, line 336:
```suggestion
\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files
\t\t\t\t\t\t\t• Can lead to false positives or upload errors when glob string is not surrounded by quotation marks`,
`,
```

README.md
```diff
@@ -705,6 +707,15 @@ Users can perform a bulk upload by using the upload-file command on a target fol
...
 
+### Uploading files using a glob pattern (BETA)<a id="glob-upload"></a>
+
+Users can declare a glob pattern to select files to be uploaded in bulk. The `upload-file` command will parse the glob pattern and will upload all the files that match the pattern to a specified parent folder.
+The example below will upload all `jpg` files from the indicated path.
+
+```shell
+ardrive upload-file --glob "/path/to/photos/folder/*.jpg" --parent-folder-id "9af694f6-4cfc-4eee-88a8-1b02704760c0" -w /path/to/wallet.json
+```
+
```
arielmelendez, line 718:
```suggestion

NOTE: Ensure that the glob pattern is surrounded by quotations marks or unexpected upload behaviors or errors may occur.
```

arielmelendez (approved):

README.md
karlprieb: 👍

src/parameter_declarations.ts
karlprieb: 👍

matibat (changes requested):
I'm getting a compilation error on my side
```shell
$ yarn && yarn build
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 0s 297ms
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed in 1s 888ms
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed in 0s 337ms
➤ YN0000: Done in 2s 686ms
src/parameter_declarations.ts:338:1 - error TS1136: Property assignment expected.

338 `,
    ~~
339  forbiddenConjunctionParameters: [LocalFilesParameter, LocalFilePathParameter]
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
340 });
    ~~~
341 
    

src/parameter_declarations.ts:341:1 - error TS1160: Unterminated template literal.

341 
    


Found 2 errors.
```

src/parameter_declarations.ts
```diff
@@ -325,3 +327,14 @@ Parameter.declare({
...
+Parameter.declare({
+	name: GlobParameter,
+	aliases: ['-g', '--glob'],
+	description: `(BETA) a glob pattern to select files to be uploaded
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files-path
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files
+\t\t\t\t\t\t\t• Can lead to false positives or upload errors when glob string is not surrounded by quotation marks`,
+`,
+	forbiddenConjunctionParameters: [LocalFilesParameter, LocalFilePathParameter]
+});
```
matibat:
```suggestion

Parameter.declare({
	name: GlobParameter,
	aliases: ['-g', '--glob'],
	description: `(BETA) a glob pattern to select files to be uploaded
\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files-path
\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files
\t\t\t\t\t\t\t• Can lead to false positives or upload errors when glob string is not surrounded by quotation marks`,
	forbiddenConjunctionParameters: [LocalFilesParameter, LocalFilePathParameter]
});
```

src/commands/upload_file.ts
```diff
@@ -69,14 +86,31 @@ new CLICommand({
...
 				});
+
 				return fileParameters;
 			}
 
+			const parentFolderId: FolderID = parameters.getRequiredParameterValue(ParentFolderIdParameter, EID);
+			const globValue = parameters.getParameterValue(GlobParameter);
+
+			if (globValue) {
+				const files = glob.sync(globValue);
```
matibat, line 97: Is there a reason why we don't use the standard shell glob functionality?

We could get rid of this dep if the only stuff it is doing is resolving the glob into an array of paths. That's what the shell currently does

matibat: I'll clone this stuff and hack it locally to see if that's possible

karlprieb:
@matibat It was a typo from the latest commit fixing the parameter description. I fixed it, can you try again?
Thank you!

src/parameter_declarations.ts
matibat: It was resolved

karlprieb: @matibat It was a typo from the latest commit fixing the parameter description. I fixed it, can you try again?
Thank you!

src/commands/upload_file.ts
karlprieb: We could. Each shell can have different rules for glob and using node-glob just create a standard.
I think it's just a different approach, I don't know if we want a standard way of glob or not.
Another thing is that I'm not sure if all shells will return a list of paths.

karlprieb: For example, my current zsh configuration will only return the first path.

karlprieb: Ok, found something :D
If we change the parameter type to array shell glob will work and we don't need node-glob.
So every shell can have different glob rules, and users will need to know how to use them. It's the same for node-glob, the difference is that we have a standard to give the user if they need.
I also saw that every shell glob should return a list of paths, so I think we're good if we wanto to go with shell glob instead of node-glob. I tried with default glob rules of zsh, bash, dash and fish, which I think are the most popular ones.

image

matibat: See here: https://github.com/karlprieb/ardrive-cli/pull/1

matibat: The latest changes are moved to https://github.com/ardriveapp/ardrive-cli/pull/185 for convenience

karlprieb:
Closed in favor of #185

PR #178, opened 2021-12-07
PE-763: Dependency management for bundles

fedellen:
This PR bumps `arweave` to be consistent with `ardrive-core-js` and absorbs dependency updates to bundle compatibility


matibat (approved):
It builds just fine with the right core branch :+1:

PR #180, opened 2021-12-08
PE-729: BATS hello world

arielmelendez:
Introduced the first BATS test for end-to-end testing of the CLI. First test simply ensures that invoking `ardrive` without input arguments returns all the expected help text to the terminal.


bats_test/ardrive.bats
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bats
+
+@test "'ardrive' with no input arguments prints general help with all expected commands" {
+    run yarn ardrive
+    [ "${status}" -eq 0 ]
+    [[ "$output" = "Usage: ardrive [command] [command-specific options]
```
matibat, line 6: I'd rather declare the multi-line string in a separate variable so the conditional gets more clear. That way we can use the single brackets instead of double brackets (I think the multi-line string is the reason we are using `[[`);

Also, a minor detail:
```suggestion
    [[ "${output}" = "Usage: ardrive [command] [command-specific options]
```

matibat (approved):
Nice example :)

fedellen (approved):

bats_test/ardrive.bats
agsuy: Fixed that test and added some more

bats_test/base-reward/base-reward.bats
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bats
+
+# Support lib
+load '/home/node/packages/node_modules/bats-support/load.bash'
+# Assertions
+load '/home/node/packages/node_modules/bats-assert/load.bash'
+
+@test "'ardrive base-reward' outputs a number" {
```
arielmelendez, line 8:
```suggestion
@test "'ardrive base-reward' outputs a positive integer number" {
```

bats_test/get-mempool/get-mempool.bats
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bats
+
+# Support lib
+load '/home/node/packages/node_modules/bats-support/load.bash'
+# Assertions
+load '/home/node/packages/node_modules/bats-assert/load.bash'
+
+@test "'ardrive get-mempool | wc -l' outputs a number of less than 5 digits" {
```
arielmelendez, line 8: I'm not sure this test will stand the test of time, no pun intended 😬

bats_test/get-mempool/get-mempool.bats
```diff
@@ -0,0 +1,18 @@
...
+    # Needs wrapping for piping
+    run bash -c "yarn ardrive get-mempool | wc -l"
+    assert_output --regexp '^[0-9]{1,4}'
+}
+
+@test "'ardrive get-mempool' first line contains a valid TX" {
+    # Needs wrapping for piping
+    run yarn ardrive get-mempool
+    assert_line -n 1 --regexp '([^"]\w{42})'
```
arielmelendez, line 17: It would be nice if we could validate that a valid JSON array were returned. This is fine for now but brittle (i.e. in the case of an empty mempool). Also this is the regex we use for txIDs (it seems yours is only partial and potentially incorrect?):
```
/^(\w|-){43}$/
```

bats_test/help/base-reward.bats
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bats
+
+# Support lib
+load '/home/node/packages/node_modules/bats-support/load.bash'
+# Assertions
+load '/home/node/packages/node_modules/bats-assert/load.bash'
+
+@test "'ardrive base-reward -h' prints help with all expected fields" {
+    run -0 yarn ardrive base-reward -h
+    assert_line --index 0 'Usage: ardrive base-reward [options]'
```
arielmelendez, line 10: Why use a line based assertion rather than assert_output?

bats_test/help/base-reward.bats
```diff
@@ -0,0 +1,28 @@
...
+    assert_line --index 4 '                   multiple of 2.5 would boost a 100 Winston transaction reward'
+    assert_line --index 5 '                   to 250 Winston.'
+    assert_line -n 6 '  -h, --help       display help for command'
+}
+
+@test "'ardrive base-reward --help' prints help with all expected fields" {
+    run -0 yarn ardrive base-reward --help
+    assert_line --index 0 'Usage: ardrive base-reward [options]'
+    assert_line --index 1 'Options:'
+    assert_line -n 2 '  --boost <boost>  (OPTIONAL) a multiple of the base transaction mining reward'
```
arielmelendez, line 23: Not using assert_output aside, is there a reason to be intermixing --index and -n?

bats_test/help/no_command.bats
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bats
+
+@test "'ardrive' with no input arguments prints general help with all expected commands" {
+    run -0 yarn ardrive
+    [[ "$output" = "Usage: ardrive [command] [command-specific options]
```
arielmelendez, line 5: Let's factor the expected help output out to a variable to make this test more easily maintainable.

bats_test/get-mempool/get-mempool.bats
agsuy: Indeed its wrong. Typo.

bats_test/help/base-reward.bats
agsuy: Me learning the shorthand and trying it out. Both are the same

bats_test/help/base-reward.bats
agsuy: Issues with lines beginning with spaces. Its pretty annoying, lines work just fine.

bats_test/get-mempool/get-mempool.bats
agsuy: Fixed.

bats_test/get-mempool/get-mempool.bats
agsuy: ahah its fine now, and an easy fix if we need to change it.

agsuy (approved):
Approved

fedellen (approved):

PR #183, opened 2021-12-10
PE-781: consolidate flag aliases

matibat:
Simplifies the parameter aliases for paths:
- `--local-file-path` and `--dest-folder-path` are consolidated to `--local-path`
- `--local-files` is renamed to `--local-paths-csv`

Still pending: the `--local-paths` parameter at https://github.com/karlprieb/ardrive-cli/pull/1


matibat:
Flag renamed at #185 (`--local-paths`)
TODO: update README.md

PR #204, opened 2021-12-16
chore(BATS): Public file tests

agsuy:
Some tests that assert existence of files.

### DIFF
--- a/bats_test/downloads/file-downloads/pre_download.sh
+++ b/bats_test/downloads/file-downloads/pre_download.sh
@@ -13,4 +13,4 @@ myFiles=(
 for publicFile in ${myFiles[@]}; do
     yarn ardrive download-file -f $publicFile
 done
-exit 1
+exit 0
--- a/bats_test/downloads/file-downloads/pre_download.sh
+++ b/bats_test/downloads/file-downloads/pre_download.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+myFiles=(
+    "290a3f9a-37b2-4f0f-a899-6fac983833b3"
+    "0dc95148-ac62-4a35-bb91-3e6ac74a54a9"
+    "fa86007b-5534-4ab3-b21c-5b6ce77acb0f"
+    "d3399761-ece8-4404-b560-6822472c9f62"
+    "366b951f-c3cd-4136-b086-faecd7872394"
+    "1571ccbd-7be8-4040-98df-0e36f8ffacc9"
+    "438f4c26-0e0d-42d1-b821-11109170a794"
+)
+
+for publicFile in ${myFiles[@]}; do
+    yarn ardrive download-file -f $publicFile
+done
+exit 1
--- a/bats_test/downloads/file-downloads/public_files.bats
+++ b/bats_test/downloads/file-downloads/public_files.bats
@@ -15,25 +15,26 @@ setup_file() {
 }
 
 @test 'Verifies "1 space test.zip" file exists' {
-    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/1 space test.zip"
+    assert_exist "$(pwd)/1 space test.zip"
 }
 @test 'Verifies "arconnect-archive-team_-_ardrive.html" file exists' {
-    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/arconnect-archive-team_-_ardrive.html"
+    assert_exist "$(pwd)/arconnect-archive-team_-_ardrive.html"
 }
 @test 'Verifies "13" file exists' {
-    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/13"
+    assert_exist "$(pwd)/13"
 }
 @test 'Verifies "Biochar-sample.pdf" file exists' {
-    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/Biochar-sample.pdf"
+    assert_exist "$(pwd)/Biochar-sample.pdf"
 }
 @test 'Verifies "Coffee_chart.jpg" file exists' {
-    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/Coffee_chart.jpg"
+    assert_exist "$(pwd)/Coffee_chart.jpg"
 }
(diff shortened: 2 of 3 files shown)


bats_test/downloads/file-downloads/publlic_files.bats
```diff
@@ -0,0 +1,39 @@
...
+@test 'Verifies "Coffee_chart.jpg" file exists' {
+    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/Coffee_chart.jpg"
+}
+@test 'Verifies "samplee.mkv" file exists' {
+    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/samplee.mkv"
+}
+@test 'Verifies "test!_·$%d\%&·_.txt" file exists' {
+    #BATS parsing issue with that filename, always fails
+    skip
+    assert_exist "/home/node/ardrive-cli/bats_test/downloads/file-downloads/test!_·\$%d\%&·_.txt"
```
matibat, line 38: If the issue here is regarding the special characters of the file name, then using single quotes would do the trick:
```suggestion
    assert_exist '/home/node/ardrive-cli/bats_test/downloads/file-downloads/test!_·\$%d\%&·_.txt'
```

bats_test/downloads/file-downloads/publlic_files.bats
```diff
@@ -0,0 +1,39 @@
...
+# File methods. DEPENDS on SUPPORT lib.
+load '/home/node/packages/node_modules/bats-file/load.bash'
+# Assertions.  DEPENDS on SUPPORT lib.
+load '/home/node/packages/node_modules/bats-assert/load.bash'
+
+DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")" >/dev/null 2>&1 && pwd)"
+
+setup_file() {
+    run $DIR/pre_download.sh
+    echo "##### Downloaded 7 files" >>/home/node/bats.log
```
matibat, line 14: What log should we dump here? Custom log messages?

bats_test/downloads/file-downloads/pre_download.sh
```diff
@@ -0,0 +1,16 @@
...
+    "d3399761-ece8-4404-b560-6822472c9f62"
+    "366b951f-c3cd-4136-b086-faecd7872394"
+    "1571ccbd-7be8-4040-98df-0e36f8ffacc9"
+    "438f4c26-0e0d-42d1-b821-11109170a794"
+)
+
+for publicFile in ${myFiles[@]}; do
+    yarn ardrive download-file -f $publicFile
+done
+exit 1
```
matibat, line 16: Why'd to exit with `1` here? Shouldn't it return `0`?

bats_test/downloads/file-downloads/publlic_files.bats
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bats
+
+# Support lib
+load '/home/node/packages/node_modules/bats-support/load.bash'
+# File methods. DEPENDS on SUPPORT lib.
+load '/home/node/packages/node_modules/bats-file/load.bash'
+# Assertions.  DEPENDS on SUPPORT lib.
+load '/home/node/packages/node_modules/bats-assert/load.bash'
+
+DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")" >/dev/null 2>&1 && pwd)"
```
matibat, line 10: This line is:

 - `cd` into the parent of `${BATS_TEST_FILENAME}`
   - the `stdout` of `cd` is redirected to `/dev/null`
   - the `stderr` is redirected to `stdout`
     - if there was an error doing `cd`, then returns `pwd` to put in `DIR`
     - if `cd` had success, then that's the value to put in `DIR`
-------------------
So, is there a case where `${BATS_TEST_FILENAME}` has a wrong value?

matibat: Can't we have a hardcoded `${BATS_TEST_BASENAME}`?

bats_test/downloads/file-downloads/publlic_files.bats
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bats
```
matibat, line 1: **TODO** rename this file to `public_files.bats`

matibat (commented):
It's such a good example!

Left a couple of comments

bats_test/downloads/file-downloads/publlic_files.bats
agsuy: This way the test ll always work as it fetches the absolute path dynamically

bats_test/downloads/file-downloads/publlic_files.bats
agsuy: This is how you create log files according to BATS docs, yes

bats_test/downloads/file-downloads/publlic_files.bats
agsuy: Fixed

bats_test/downloads/file-downloads/pre_download.sh
agsuy: True. Fixed.

bats_test/downloads/file-downloads/public_files.bats
```diff
@@ -0,0 +1,40 @@
...
+
+DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")" >/dev/null 2>&1 && pwd)"
+
+setup_file() {
+    run $DIR/pre_download.sh
+    echo "##### Downloaded 7 files" >>/home/node/bats.log
+}
+
+@test 'Verifies "1 space test.zip" file exists' {
+    assert_exist "$(pwd)/1 space test.zip"
```
matibat, line 18: Shouldn't we use the `DIR` variable here?
```suggestion
    assert_exist "${DIR}/1 space test.zip"
```

agsuy: Noup. Stuff is downloaded in an external script whereas test asserts relative to where it is located.

matibat:
Approved. My only concern is we have a dynamic path for the files being downloaded. I'd like to see them in the same folder always, but tet's discuss that behavior later on

PR #209, opened 2021-12-23
PE-459: Single file bundles

fedellen:
This PR adds support from Core for uploading single files as a bundle transaction.


bats_test/upload-file/single-bundled-files.bats
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bats
+
+# Support lib
+load '/home/node/packages/node_modules/bats-support/load.bash'
+# Assertions
+load '/home/node/packages/node_modules/bats-assert/load.bash'
+
+setup_file() {
+    run $DIR/create-file.sh
+    echo "##### Created 10 chunk test file" >>/home/node/bats.log
```
agsuy, line 10: Can we create a couple more files on this script to test with different chunk sizes?
Its a cheap way to greatly increase coverage

bats_test/upload-file/upload-test.sh
```diff
@@ -2,6 +2,6 @@
 
 cd /home/node/ardrive-cli
 
-yarn ardrive upload-file --dry-run -w $WALLET -l '/home/node/10Chunks.txt' -F $PUB_FOLD_ID | jq -c '.fees' | awk -F '"' '{ print $4 }'
+yarn ardrive upload-file --no-bundle --dry-run -w $WALLET -l '/home/node/10Chunks.txt' -F $PUB_FOLD_ID | jq -c '.fees' | awk -F '"' '{ print $4 }'
```
agsuy, line 5: Why --no-bundle here?

agsuy (commented):
Left 2 comments that could lead to changes

bats_test/upload-file/upload-test.sh
fedellen: Bundling the transaction results in a reward for 11 chunks, which causes the test to fail

jdaev (approved):

src/commands/upload_file.ts
```diff
@@ -156,11 +160,13 @@ new CLICommand({
 			const wallet = parameters.getRequiredParameterValue(WalletFileParameter, readJWKFile);
 
 			const conflictResolution = parameters.getFileNameConflictResolution();
+			const shouldBundle = !!parameters.getParameterValue(NoBundleParameter);
```
matibat, line 163: ~Is it `shouldBundle` or `shouldNotBundle`?~
- ~if `--no-bundle`, then `parameters.getParameterValue(NoBundleParameter)` is true~
- ~`!!true` is true~

**EDIT**
This is fine because the parameter declarations specify the `bundle` parameter name, but `--no-bundle` is the alias so the negation is already applied.
Maybe the confusing thing here is the name of the parameter. It should be `ShouldBundleParameter` or `BundleParameter`

matibat (approved):
Left one single comment. It seems good overall!

matibat (dismissed):
LGTM

bats_test/upload-file/single-bundled-files.bats
```diff
@@ -0,0 +1,48 @@
...
+
+    assert_line -n 0 '"file"'
+    assert_line -n 1 ''
+}
+
+@test "upload bundled file produces one fee" {
+    run -0 bash -c "yarn ardrive upload-file --dry-run -l '/home/node/10Chunks.txt' -F $PUB_FOLD_ID -w $WALLET | jq '.fees'"
+
+    assert_line -n 0 '{'
+    assert_line -n 1 --regexp '(\s){0,}("){1}(\w|-){43}("){1}'
```
arielmelendez, line 37: You could save yourself the some of the pageantry of the regex by using jq in raw mode (gets rid of the whitespace and quotes), getting the fee keys, and unpacking the output array , e.g.:
```
| jq -r '.fees | keys | .[]'
```
yielding output like:
```
7DPiA3gv0Hcsn8ILpYIhcOYvV0udjyUnMs7G3bDrLhI
```

arielmelendez (dismissed):

bats_test/upload-file/single-bundled-files.bats
fedellen: Added: 34ca7f3

package.json
```diff
@@ -8,7 +8,7 @@
 	},
 	"types": "./lib/index.d.ts",
 	"dependencies": {
-		"ardrive-core-js": "1.4.0",
+		"ardrive-core-js": "1.5.0-alpha-3",
```
fedellen, line 11: TODO: Use prod release

arielmelendez (approved):

PR #219, opened 2022-01-05
Pe 832 support drives folders

agsuy:
Only Docs and one BASH script. Interaction is with BATS docker only.

- Migrates documentation form Docker repository
- Adds new folder to hold BATS scripts

### DIFF
--- a/.bats/entities.sh
+++ b/.bats/entities.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+cd $HOME/ardrive-cli/
+
+cache="$(yarn ardrive list-all-drives -w $WALLET)"
+driveID="$(echo $cache | jq '[.. | select(.drivePrivacy? == "public")][0] | .driveId' | cut -d '"' -f 2)"
+folderID="$(echo $cache | jq '[.. | select(.drivePrivacy? == "public")][0] | .rootFolderId' | cut -d '"' -f 2)"
+
+export PUB_DRIVE_ID="$driveID"
+export PUB_FOLD_ID="$folderID"
--- a/bats_test/network_tools.md
+++ b/bats_test/network_tools.md
@@ -1,0 +1,35 @@
+# Network tests
+
+Useful to monitor both inbound and outbound traffic to BATS Docker
+
+Disclaimer: *Might now work on MacOS*
+
+With your ardrive-bats-docker running, in another terminal we run the following command:
+
+```docker run -it --rm --cap-add=NET_ADMIN --net container:ardrive-cli-bats nicolaka/netshoot```
+
+This will open a new container using the public Netshoot image that controls CLI docker network capabilities.
+
+To see a list of every included package as well as some examples please check [Netshoot repo](https://github.com/nicolaka/netshoot#netshoot-a-docker--kubernetes-network-trouble-shooting-swiss-army-container)
+
+
+## Usage examples
+### Redirecting Traffic
+
+To "disable" any host, we just need to redirect its traffic.
+
+```echo "{IP where we want to redirect} {host I want to redirect}" >> /etc/hosts```
+
+A real example would be ```echo "0.0.0.0 http://ardrive.io" >> /etc/hosts``` to redirect all the traffic to an invalid IP (local-host)
+
+In order to mimic/achieve different behaviors, we can use ```iproute2``` 
+
+e.g. to get an "Unreachable" we could run this command inside Netshoot image
+
+```ip route add unreachable <IP we redirected>```
+
+For more examples please check [iproutes2 documentation](https://baturin.org/docs/iproute2/#ip-route-add-blackhole)
+
+### Measure traffic
+
+Inside Netshoot, run ```iftop``` to monitor connections and measure speeds e.g. while uploading/downloading
--- a/bats_test/readme.md
+++ b/bats_test/readme.md
@@ -6,7 +6,7 @@ To run a single file, just use
 
 e.g.
 
-```bats ../test_samples//testing_hooks/hooks_sample.bats```
+```bats ../test_samples/testing_hooks/hooks_sample.bats```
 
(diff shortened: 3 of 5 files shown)


.bats/entities.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+cd $HOME/ardrive-cli/
+
+cache="$(yarn ardrive list-all-drives -w $WALLET)"
+driveID="$(echo $cache | jq '[.. | select(.drivePrivacy? == "public")][0] | .driveId' | cut -d '"' -f 2)"
+folderID="$(echo $cache | jq '[.. | select(.drivePrivacy? == "public")][0] | .rootFolderId' | cut -d '"' -f 2)"
```
matibat, line 7: I tested it in isolation and works well. Although I'd like to get rid of that optional field (i.e. `?`) so we enforce the input to be in good shape
```suggestion
driveID="$(echo $cache | jq --raw-input '[.[] | select(.drivePrivacy == "public")][0] | .driveId')"
folderID="$(echo $cache | jq --raw-input '[.[] | select(.drivePrivacy == "public")][0] | .rootFolderId')"
```
**Explanation**
1. removing the question mark throws an error (cannot index an array with a string value)
2. replacing the recursive operator by explicitly iterating the expected array. `..` -> `.[]` 
3. we get rid of the call to `cut` to remove the quotes by using the `-r --raw-input` flag at `jq`

.bats/entities.sh
```diff
@@ -0,0 +1,10 @@
...
+
+export PUB_DRIVE_ID="$driveID"
+export PUB_FOLD_ID="$folderID"
```
matibat, line 10: We could replace the above variable declarations for two functions `get_1st_drive_id_from_wallet` and `get_1st_folder_id_from_wallet`. That way we could fancily do the following
```suggestion
export PUB_DRIVE_ID="$(get_1st_drive_id_from_wallet ${WALLET})"
export PUB_FOLD_ID="$(get_1st_folder_id_from_wallet ${WALLET})"
```
But again: it's just to be fancy :nail_care:

bats_test/network_tools.md
```diff
@@ -0,0 +1,35 @@
+# Network tests
+
+Useful to monitor both inbound and outbound traffic to BATS Docker
+
+Disclaimer: *Might now work on MacOS*
```
matibat, line 5: Why so? :/

bats_test/readme.md
```diff
@@ -1,64 +1,33 @@
...
-Some guidelines on writing your own BATS tests
-## Globals
+To run a single file, just use
+``` bats <my-test-file.bats>```
 
-There are several global variables you can use to introspect on Bats tests:
+e.g.
 
-- $BATS_TEST_FILENAME is the fully expanded path to the Bats test file.
+```bats ../test_samples//testing_hooks/hooks_sample.bats```
```
matibat, line 9: Nit
```suggestion
```bats ../test_samples/testing_hooks/hooks_sample.bats```
```

bats_test/wallets.md
```diff
@@ -0,0 +1,39 @@
+# Interact with a wallet
+
+
+
```
matibat, line 4: nit
```suggestion

```

bats_test/wallets.md
```diff
@@ -0,0 +1,39 @@
...
+## Wallet Operations
+
+There is a $WALLET variable directly pointing to /home/node/tmp/wallet.json inside the Docker.
+
+In order to run any command that requires a wallet you could just replace its path with $WALLET
+
+e.g. for a private file
+
+``yarn ardrive file-info -f [file-id] -w $WALLET -p [my-unsafe-password]``
```
matibat, line 39: The syntaxis of `[...]` in the shell context usually refers to an optional and might be confusing. Consider using an example variable name instead.
```suggestion
``yarn ardrive file-info -f $FILE_ID -w $WALLET -P``
```

matibat: Also, please confirm that the two backticks are what we want. I usually see one or three

bats_test/writing_tests.md
```diff
@@ -0,0 +1,64 @@
...
+[Files](https://github.com/bats-core/bats-file#index-of-all-functions)
+
+[Assertions](https://github.com/bats-core/bats-assert#usage)
+
+
+## Basic Commands
+
+### Log to console
+
+Always  ``` echo 'text' >&3 ```
```
matibat, line 38: Would you mind specifying which STD stream does `>&3` stand for? Is it the error stream?

bats_test/writing_tests.md
```diff
@@ -0,0 +1,64 @@
...
+
+Only on failures ```echo 'text' ```
```
matibat, line 40: Something special happens when I log an error like that?

matibat (dismissed):
Please look my comments

bats_test/writing_tests.md
agsuy: One is only on failures, the other always logs

bats_test/writing_tests.md
agsuy: Its on BATS documentation. Its a way to log directly https://bats-core.readthedocs.io/en/stable/writing-tests.html#file-descriptor-3-read-this-if-bats-hangs

bats_test/network_tools.md
agsuy: Untested. 
Networking on Mac is a bit limited so I cannot assure everything works there. 
https://docs.docker.com/desktop/mac/networking/

.bats/entities.sh
agsuy: This is opinionated, gonna keep shorter names for now.

arielmelendez (approved):

agsuy:
We could iterate on jq command on a follow up but it is not a priority.

PR #224, opened 2022-01-12
PE-460: Bulk bundles

fedellen:
This PR uses the new uploadAllEntities method from Core for all upload paths. The expected results for a bundled multi file upload have been changed to expect one combined bundle. New BATS test cases have been added for a bulk folder


src/commands/upload_file.ts
```diff
@@ -169,66 +169,29 @@ new CLICommand({
...
-									prompts: fileUploadConflictPrompts
-								});
-							}
-						}
-					})();
-				})
+			const uploadStats: ArDriveUploadStats[] = await Promise.all(
+				filesToUpload.map(
+					async ({ parentFolderId, wrappedEntity, destinationFileName, driveKey, drivePassword }) => {
+						driveKey ??= (await parameters.getIsPrivate())
```
matibat, line 175: I assume that this operator is conditionally setting the value only if null. There are a couple of places in the codebase where this would spare us an `if` and thus one indentation level. Nice approach!

matibat (dismissed):
I like how the changes at CORE made this code simpler by using that method which doesn't care about the entities' privacy, but instead uses a generic type :clap:

matibat:
Nit: there's a conflict at the README.md

src/commands/upload_file.ts
fedellen: Yes, that is the case: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment

fedellen:
> I like how the changes at CORE made this code simpler by using that method which doesn't care about the entities' privacy, but instead uses a generic type 👏

We needed to abstract the privacy away from the core method here because CSV uploads can be sent up to different destination folders with potentially different privacy levels

bats_test/upload-file/create-file.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-yes "10 chunk test"  | head --bytes 2621440 > /home/node/10Chunks.txt
+yes "10 chunk test"  | head -c 2621440 > /home/node/10Chunks.txt
```
arielmelendez, line 3: Is `-c` portable across OSes? I know --bytes doesn't work on OSX but I'm not sure if -c works on Linux (I may have that backwards, but double check either way).

src/commands/upload_file.ts
```diff
@@ -169,66 +169,29 @@ new CLICommand({
...
+						return { wrappedEntity, driveKey, destFolderId: parentFolderId, destName: destinationFileName };
+					}
+				)
 			);
 
-			console.log(JSON.stringify(formatResults(results), null, 4));
+			const results = await arDrive.uploadAllEntities({
+				entitiesToUpload: uploadStats,
+				conflictResolution,
+				prompts: folderUploadConflictPrompts
```
arielmelendez, line 191: Nit: This name always throws me off. `fileOrFolderUploadConflictPrompts` would likely make more sense since they're composed together.

arielmelendez (approved):

bats_test/upload-file/create-file.sh
fedellen: Not sure about Windows, but -c works on inside our Linux Docker environment and on MacOS. In hindsight, this change was unnecessary. I was just having local compatibility issues there

src/commands/upload_file.ts
fedellen: Changed that one over to `fileAndFolderUploadConflictPrompt`, since they contain both: 6c14939

PR #228, opened 2022-01-17
PE-894: Bundle readme updates

fedellen:


README.md
```diff
@@ -17,12 +17,15 @@ ardrive create-drive --wallet-file /path/to/my/wallet.json --drive-name "Teenage
...
+        {
+            "type": "bundle",
+            "bundleTxId": "Vj2x4IBEAezBvhj5RgtA247W_q3S10suI6l0E30GPoE"
         }
     ],
     "tips": [],
     "fees": {
-        "giv2R8Xj0bbe6l5taBTQJk_38zwIrMH_g1-knSCisjU": 1415103,
-        "VljnttwUxRStnVuPYakF9e2whjhYJVWB0nSxD5dVyJ8": 1391904
+          "Vj2x4IBEAezBvhj5RgtA247W_q3S10suI6l0E30GPoE": "44579472"
```
arielmelendez, line 28: Nit: Tabbing looks inconsistent in the diff on GH

README.md
```diff
@@ -34,21 +37,23 @@ ardrive upload-file --wallet-file /path/to/my/wallet.json --parent-folder-id "f0
...
                 "address": "i325n3L2UvgcavEM8UnFfY0OWBiyf2RrbNsLStPI73o"
             },
             "winston": "10000000"
         }
     ],
     "fees": {
-        "tSMcfvAQu_tKLUkdvRRbqdX93oAf3h6c9eJsSj8mXL4": 1384601,
-        "EvE06MmE9IKeUzFMnxSgY1M5tJX4uHU64-n8Pf_lZfU": 1447752,
-        "FidEhcZtmDtvQxWrnVJlKnj_ZkwxYXvn7wjbUpasRKo": 1379016
+        "qjdHiQoWlSjCvhj5RgtA247W_q3S10suI6l0E30GPoE": 44579472,
```
arielmelendez, line 56: Invalid trailing comma in JSON

README.md
```diff
@@ -733,7 +743,7 @@ ardrive upload-file -w wallet.json -F "6939b9e0-cc98-42cb-bae0-5888eca78885" --l
 ardrive upload-file -w wallet.json -F "6939b9e0-cc98-42cb-bae0-5888eca78885" --local-paths ./*.json
 ```
 
-### Name Conflict Resolution on Upload
+### Name Conflict Resolution on Upload<a id="conflict-resolution"></a>
```
arielmelendez, line 746: Is anything using the `conflict-resolution` anchor yet?

fedellen: Nope, I noticed it was missing and meant to add that here

fedellen: 57562f5

README.md
```diff
@@ -781,6 +792,36 @@ Please select how to proceed:
...
+
+Bundles do however come with their own limitations. Currently, any file data that is larger than 500 MiB will not be bundled and will instead be uploaded as a v2. Also, bundles will never contain more than 500 data items. These limitations have been chosen to avoid unpacking issues experienced at the gateway.
+
+### Bundled Transactions Impact on Upsert Option and List Commands<a id="bundle-impact"></a>
+
+Another major limitation for bundled transactions is how long they can take to be unpacked by the gateway and become available using a GQL query. With v2 transactions, the gateway will typically optimistically index within a few seconds. But with bundled transactions, they must first be unpacked in order to be indexed. The length of this process will greatly fluctuate based on current gateway health and conditions, but it typically takes between 10-60 minutes.
+
+Unfortunately, this has a negative impact on the CLI's user experience in it's current state. As a one time executable node program with no persistence layer, the CLI will have no knowledge that your ArFS entities inside of bundled transactions exist until they are available from the gateway via a GQL query.
```
arielmelendez, line 809:
```suggestion
Unfortunately, this has a negative impact on the CLI's user experience in its current state. As a one-time executable node program with no persistence layer, the CLI will have no knowledge that your ArFS entities inside of bundled transactions exist until they are available from the gateway via a GQL query.
```

README.md
```diff
@@ -781,6 +792,36 @@ Please select how to proceed:
     Skip this file upload
 ```
 
+### Understanding Bundled Transactions vs V2 Transactions<a id="bundle-vs-v2"></a>
```
arielmelendez, line 795: I think it could be beneficial to reduce the amount of information conveyed in these sections to get the following key points across to the reader quickly:
```
• Bundling saves AR and enhances ArFS reliability by sending associated ArFS transactions up as one atomic bundle.
• Bundled transactions are treated as a single data transaction by the Arweave network, but can be presented as separate transactions by the Arweave Gateway once they have been "unbundled".
• Unbundling can take anywhere from a few minutes up to an hour. During that time, the files in the bundle will neither appear in list- commands nor be downloadable. Similarly, they will not appear in the web app after syncs until unbundling is complete. This can negatively affect the accuracy of upsert operations, so it's best to wait before retrying bulk uploads.
• Bundling is the default option for the CLI, but can be disabled by using the --no-bundle option.
• Bundling reliability on the gateway side degrades once bundles reach either 500 data items (or ~250 files) or 500MiB, so the CLI will create and upload multiple bundles as necessary, or will send files that are simply too large for reliable bundling as unbundled txs.
```
Many of the points I've listed there are included in your prose below. But if we can get straight to the point without too much supplementary info, I think it could streamline uptake for the user. For example, I don't think it's critically important for users to know about V2 transactions, particularly by that proper name, unless it's a term we need to use often to convey how to use the CLI correctly.

fedellen: Sounds good, I've added your suggestion here: 684fe92

Let me know if I've taken too much away

jdaev (approved):

matibat (approved):
LGTM

arielmelendez (approved):

PR #233, opened 2022-01-25
Pe 869 create bats tests single file bundles

agsuy:
- New script to load entities.
- Updated documentation
- New tests (2) using new variables for bundles testing conflict resolution


bats_test/upload-file/single-bundled-files.bats
```diff
@@ -60,4 +59,22 @@ setup_file() {
...
+    #We use quotes to escape possible spaces
+    run bash -c "touch '$PUB_FILE_NAME'"
+    #Again we use quotes to escape spaces
+    run -0 bash -c "yarn ardrive upload-file --dry-run --local-path ./'$PUB_FILE_NAME' -F $PARENT_FOLDER_ID -w $WALLET --skip | jq -r '.created | length'"
+
+    assert_output 0
+}
+
+@test "Duplicate name uploads a new file with --upsert" {
+    run bash -c "touch $PUB_FILE_NAME"
```
matibat, line 75: nit
```suggestion
    run bash -c "touch '$PUB_FILE_NAME'"
```

matibat: Beware the single quotes are working fine because the variables are addressed before passing the string to `bash`. If we weren't using bash here, then we must be using double-quotes.

matibat (dismissed):
Left a nit comment. Looking good overall! :clap:

matibat (approved):

PR #235, opened 2022-01-26
PE-897: BATS for rename-folder

matibat:


.bats/entities.sh
```diff
@@ -8,29 +8,61 @@ cd $HOME/ardrive-cli/
...
-        fileSize="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .size')"
-        fileName="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .name')"
-        fileNameAlt="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][1] | .name')"
-        folderID="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .parentFolderId')"
-        rootID="$(echo $aux | jq -r '[.[] | select(.parentFolderId == "root folder")][0] | .entityId')"
-        driveID="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .driveId')"
+found=0
+for publicDriveId in ${allDriveIDs[@]}; do
+    driveContent="$(yarn ardrive list-drive -d $publicDriveId)"
+    # numberFiles="$(echo $driveContent | jq -r '[.[] | select(.entityType == "file")] | length')"
```
matibat, line 17: remove commented code

.bats/entities.sh
```diff
@@ -8,29 +8,61 @@ cd $HOME/ardrive-cli/
...
+    allFolderIDs="$(echo $driveContent | jq -rc '.[] | select(.entityType == "folder") | .entityId')"
+
+    for someFolderId in ${allFolderIDs[@]}; do
+        childFileIDs=($(echo $driveContent | jq -rc ".[] | select((.entityType == \"file\") and .parentFolderId == \"${someFolderId}\") | .entityId"))
+        childFolderIDs=($(echo $driveContent | jq -rc ".[] | select((.entityType == \"folder\") and .parentFolderId == \"${someFolderId}\") | .entityId"))
+
+        printf "Iterating folder with ID: ${someFolderId} which cointains ${#childFileIDs[@]} files and ${#childFolderIDs[@]} folders\n"
+        printf "chilfFiles=${childFileIDs[@]}\n"
+        printf "chilfFolders=${childFolderIDs[@]}\n"
```
matibat, line 26: remove debug prints

matibat:
```suggestion
```

.bats/entities.sh
matibat:
```suggestion
```

.bats/entities.sh
```diff
@@ -8,33 +8,61 @@ cd $HOME/ardrive-cli/
...
-        fileID="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .entityId')"
-        fileSize="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .size')"
-        fileName="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .name')"
-        fileNameAlt="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][1] | .name')"
-        folderID="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .parentFolderId')"
-        rootID="$(echo $aux | jq -r '[.[] | select(.parentFolderId == "root folder")][0] | .entityId')"
-        driveID="$(echo $aux | jq -r '[.[] | select(.entityType == "file")][0] | .driveId')"
+    allFolderIDs="$(echo $driveContent | jq -rc '.[] | select(.entityType == "folder") | .entityId')"
+
+    for someFolderId in ${allFolderIDs[@]}; do
```
arielmelendez, line 23: All of this is getting messy and complicated. We really need to get ArLocal up and running soon to create data sets on the fly.

arielmelendez: I think my chief complaint is that there's not a simple explanation in a comment up above about what we're trying to accomplish here. The script description mentions `sets ENV variables using the values from the first valid Public file` which is not only partially valid. And also there's a whole workflow here dedicated to figuring out a structure, but that's not explicitly spelled out above. So I end up needing to read a bunch of code and conditionals to understand what you're trying to find.

bats_test/rename/rename-folder/rename_colliding_name.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Depends on ./rename_folder.sh
+
+rename_colliding_name() {
```
arielmelendez, line 5: Would it make more sense to let this file take an input argument for the old file name and then just use it as the input parameter to ardrive? Also, it looks like you're missing the -n flag.

arielmelendez (approved):
Approved though entities.sh is getting overbloated. ArLocal FTW.

bats_test/rename/rename-folder/rename_public_folder.bats
```diff
@@ -0,0 +1,71 @@
...
+# Support lib
+load '/home/node/packages/node_modules/bats-support/load.bash'
+# File methods. DEPENDS on SUPPORT lib.
+load '/home/node/packages/node_modules/bats-file/load.bash'
+# Assertions.  DEPENDS on SUPPORT lib.
+load '/home/node/packages/node_modules/bats-assert/load.bash'
+
+# The function which builds and runs the command
+load './rename_folder.sh'
+# A function which triggers the rename with invalid names. DEPENDS on rename_file.sh
```
karlprieb, line 12:
```suggestion
# A function which triggers the rename with invalid names. DEPENDS on rename_folder.sh
```
:smile:

bats_test/rename/rename-folder/rename_public_folder.bats
```diff
@@ -0,0 +1,71 @@
...
+load './rename_invalid_names.sh'
+# A function which triggers a rename with an already in use name. DEPENDS on rename_file.sh
```
karlprieb, line 14:
```suggestion
# A function which triggers a rename with an already in use name. DEPENDS on rename_folder.sh
```

bats_test/rename/rename-folder/rename_public_folder.bats
```diff
@@ -0,0 +1,71 @@
...
+load './rename_colliding_name.sh'
+# A function which triggers a rename with exactly the same name. DEPENDS on rename_file.sh
```
karlprieb, line 16:
```suggestion
# A function which triggers a rename with exactly the same name. DEPENDS on rename_folder.sh
```

bats_test/rename/rename-folder/rename_colliding_name.sh
matibat: If so, then I would remove the file and directly use the `rename_folder.sh` from the `.bats` file which does exactly that

BTW What this method is running here is the case in which you try to rename the folder the same way as some sibling is (a file in this case). For renaming the folder into its current name there is `rename_same_name.sh`

.bats/entities.sh
matibat: Added comments for clarification: f5cac008545cb38f1cb87a91b2859e579bba64df

arielmelendez (approved):
Left some comments on the updates. No need for more changes. Let's get this shipped.

PR #236, opened 2022-02-01
PE-967: Fix priv folder uploads

fedellen:
This PR will apply the fix from Core for private folder uploads containing files

GH Release Notes:

- Bulk folder uploads to a private drive will no longer throw errors


matibat (approved):
LGTM

PR #239, opened 2022-02-03
PE-979: Fix folder upload contents target

fedellen:
This PR merges in the changes from Core to fix folder uploads.

Proposed GitHub Release Notes:

- File contents of a folder upload will now correctly be uploaded inside of that folder on chain


matibat (approved):
Correct version bumps for both the app version and the core dependency :+1:

karlprieb (approved):
👍

PR #243, opened 2022-02-07
PE-994: Release v1.8.0

matibat:
Release notes:
- Adds the renaming functionality: rename-file, rename-folder and rename-drive.
- Retrieve entity data from the gateway's optimistic cache and retry up to 5 times if necessary.


fedellen (approved):
Good to go!

PR #246, opened 2022-02-07
PE-480: Custom gateway flag

fedellen:
This PR adds the `--gateway` flag on every command that uses a gateway and also introduces an ENV variable named `ARWEAVE_GATEWAY`. The gateway is fed down to Core through the `arDriveFactory` in each case that an `arweave` instance is used.

I've made use of the built in `URL` class for some basic type safety and sanitization of user input. This implementation has a set of tests to validate that our use cases would work as expected.


bats_test/base-reward/base-reward.bats
```diff
@@ -9,3 +9,8 @@ load '/home/node/packages/node_modules/bats-assert/load.bash'
     run -0 yarn ardrive base-reward
     assert_output --regexp '^[0-9]+$'
 }
+
+@test "'ardrive base-reward --gateway' will error with an invalid Arweave gateway" {
+    run yarn ardrive base-reward --gateway http://fakeway.net
```
arielmelendez, line 14: Nit: If `fakeway.net` ever becomes real, this test could break...

arielmelendez: Same comment for all the other usages of fakeway.net in this PR.

arielmelendez, line 14: Should also check exit status. I think there's 2 ways to do that:
https://bats-core.readthedocs.io/en/stable/writing-tests.html#run-test-other-commands
1) Check the value of the $status variable.
2) Invoke `run -1` which checks that the exit status is 1.

I'm not sure what the tradeoffs are of each but the second option looks pretty clean.

arielmelendez: Same comment for all other examples of this in the PR.

src/CLICommand/parameters_helper.ts
```diff
@@ -277,6 +281,37 @@ export class ParametersHelper {
 		return mapFunc(value);
 	}
 
+	// Parse user input through `URL` constructor. This will throw an error on any
+	// provided url that does not have a `protocol` and `host` or is otherwise invalid
+	public getGateway(): URL {
```
arielmelendez, line 286: Consider renaming to `getOptionalGateway` to keep with the convention we've established elsewhere (e.g. getOptionalWallet).

src/CLICommand/parameters_helper.ts
```diff
@@ -277,6 +281,37 @@ export class ParametersHelper {
...
+
+			return undefined;
+		})();
+
+		if (!userProvidedURL) {
+			// Return default CLI Arweave if no gateway can be derived from the user
+			return new URL(DEFAULT_GATEWAY);
+		}
+
+		if (userProvidedURL.hostname === '') {
```
arielmelendez, line 308: What case is this?

src/CLICommand/parameters_helper.ts
```diff
@@ -277,6 +281,37 @@ export class ParametersHelper {
 		return mapFunc(value);
 	}
 
+	// Parse user input through `URL` constructor. This will throw an error on any
```
arielmelendez, line 284: Might be nice to know from the docs that this could throw.

src/commands/base_reward.ts
```diff
@@ -2,20 +2,21 @@ import { ByteCount } from 'ardrive-core-js';
 import { CLICommand, ParametersHelper } from '../CLICommand';
 import { CLIAction } from '../CLICommand/action';
 import { SUCCESS_EXIT_CODE } from '../CLICommand/error_codes';
-import { BoostParameter } from '../parameter_declarations';
+import { BoostParameter, GatewayParameter } from '../parameter_declarations';
 import axios, { AxiosResponse } from 'axios';
 
-async function getBaseReward(byteCount?: ByteCount): Promise<string> {
-	const response: AxiosResponse = await axios.get(`https://arweave.net/price/${byteCount ?? 0}`);
```
arielmelendez, line 9: Double check for all references in CLI code to https://arweave.net.

src/commands/create_manifest.ts
```diff
@@ -25,17 +27,20 @@ new CLICommand({
...
 		const parameters = new ParametersHelper(options, cliWalletDao);
 
 		const wallet = await parameters.getRequiredWallet();
+		const arweave = getArweaveFromURL(parameters.getGateway());
 
 		const arDrive = cliArDriveFactory({
 			wallet: wallet,
 			feeMultiple: parameters.getOptionalBoostSetting(),
-			dryRun: !!options.dryRun
+			dryRun: !!options.dryRun,
```
arielmelendez, line 42: should use `isDryRun` function

src/index.ts
```diff
@@ -44,6 +44,9 @@ export const cliArweave = Arweave.init({
 
 export const cliWalletDao = new WalletDAO(cliArweave, CLI_APP_NAME, CLI_APP_VERSION);
 
+export const customArweaveCliWalletDAO = (arweave = cliArweave): WalletDAO =>
```
arielmelendez, line 47: Consider a factory function-like name

arielmelendez (changes requested):
Use `-1` in BATS run commands for those you expect to fail.

src/index.ts
fedellen: [681d247](https://github.com/ardriveapp/ardrive-cli/pull/246/commits/681d247ed4c9510df8496ae3f303354500020c8a)

src/commands/create_manifest.ts
fedellen: [352bfc1](https://github.com/ardriveapp/ardrive-cli/pull/246/commits/352bfc1f91ab4ea0c59ec96cd90d89f0eb9496a9)

src/CLICommand/parameters_helper.ts
fedellen: Added docblock: [a21f4a1](https://github.com/ardriveapp/ardrive-cli/pull/246/commits/a21f4a1a0cab105eefebb9a6ac63572b730fa5f4)

bats_test/base-reward/base-reward.bats
fedellen: Updated here: [033556a](https://github.com/ardriveapp/ardrive-cli/pull/246/commits/033556a9c53f986d9e182af94521e6a908c4aa1e)

fedellen: Changed this to `http://localhost:1337`: [033556a](https://github.com/ardriveapp/ardrive-cli/pull/246/commits/033556a9c53f986d9e182af94521e6a908c4aa1e)

karlprieb (dismissed):
👍

arielmelendez (approved):

src/parameter_declarations.ts
```diff
@@ -441,3 +443,9 @@ Parameter.declare({
 		DestinationFileNameParameter
 	]
 });
+
+Parameter.declare({
+	name: GatewayParameter,
+	aliases: ['-g', '--gateway'],
+	description: `(OPTIONAL) a 'protocol://host:port' formatted string specifying the connection info for the Arweave gateway server to use`
```
matibat, line 450: Would it make sense to specify the default one as it is optional?

matibat (dismissed):
**LGTM**
Left a couple of non-blocking comments

README.md
```diff
@@ -1128,6 +1129,41 @@ Transport your signed transaction to the Internet-connected machine and run the
 ardrive send-tx -x /path/to/sendme.json
 ```
 
+### Using a Custom Arweave Gateway
```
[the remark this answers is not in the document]
fedellen: These headers always map to anchors by default. We do add custom anchors sometimes when the header titles are too verbose or for linking to areas without a header. 

It'll always map to uncapitalized kebab case; e.g `Using a Custom Arweave Gateway` has a default anchor with an id of `#using-a-custom-arweave-gateway`

karlprieb (dismissed):

arielmelendez (approved):

PR #250, opened 2022-02-12
PE-1017: remove any wrapping jwk object from generated wallet

### LINKED ISSUE #248: wallets produced by generate-wallet do not load: jwk wrapping object
it looks like generate-wallet outputs wallets wrapped by `{"jwk":}`. this produces 'undefined' errors when the output wallets are later loaded.

example:
```
$ ardrive generate-wallet -s 'this is an example twelve word seed phrase that you could use' > ardrive_wallet.json
$ ardrive get-address --wallet-file ardrive_wallet.json
TypeError: Cannot read property 'replace' of undefined
```

likely very easy to patch but i thought i'd open an issue because it seems very strange that such basic functionality isn't working. how is everybody generating wallet files?

xloem:
fixes #248


fedellen (dismissed):
Looks like a good fix at the CLI layer. Even has support for any wallets with expected shapes.

Thanks!

matibat (approved):
Excellent!

src/commands/generate_wallet.ts
```diff
@@ -12,7 +12,7 @@ new CLICommand({
 		const parameters = new ParametersHelper(options);
 		const seedPhrase = parameters.getRequiredParameterValue(SeedPhraseParameter);
 		const wallet = await cliWalletDao.generateJWKWallet(new SeedPhrase(seedPhrase));
-		console.log(JSON.stringify(wallet));
+		console.log(JSON.stringify(wallet['jwk'] || wallet));
```
arielmelendez, line 15: Remove the || and/or add a BATS test to validate our output.

arielmelendez (changes requested):
See my comment.

xloem:
I've added a BATS test. BATS is new to me.

Honestly the reason for the `||` is because I don't know whether you guys would consider the error to be in the CLI layer or the backend function.

bats_test/generate-wallet/generate-wallet.bats
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bats
```
karlprieb, line 1: Thank you for your PR @xloem 👍 

As you're adding BATS for `generate-wallet` can we add one to test when it failed?
Example:
`yarn ardrive generate-wallet -s 'invalid seed phrase'`
It should failed with the following message: `Error: 'invalid seed phrase' is not a valid 12 word seed phrase!`

arielmelendez (approved):
Thanks for adding the BATS happy path test!

xloem:
I added a test for invalid long and short seed phrases.

I also added tests for get-address on the same wallet. Unsure if the style is correct.

bats_test/get-address/get-address.bats
```diff
@@ -0,0 +1,46 @@
...
+    assert_line -n 0 "Error: 'invalid seed phrase' is not a valid 12 word seed phrase!"
+    
+    run -1 bash -c "yarn ardrive get-address -s 'this invalid seed phrase has thirteen words and is expected to be rejected'"
+
+    assert_line -n 0 "Error: 'this invalid seed phrase has thirteen words and is expected to be rejected' is not a valid 12 word seed phrase!"
+}
+
+@test "get-address finds the correct address of a canonical wallet" {
+
+    ! read -d '' WALLET <<DOC_END
```
karlprieb, line 28: You could just declare a simple WALLET variable like this:
```sh
WALLET='
{
  "kty": "RSA",
  ...
}
'
```
and use it later with:
```sh
run -0 bash -c 'echo $WALLET | yarn ardrive get-address -w /dev/stdin'
```
It's easier to read and don't depend on `read` :)

xloem: Thanks. I've made that change.

bats_test/get-address/get-address.bats
```diff
@@ -0,0 +1,46 @@
...
+        "dq": "B_WrEgZ8qY_3vLlJHsr5kJ26VqPgKiQRnJIsb6fsQdKOoUofTP3A0rzmZRAB1ltA-tChyBCMf9leBfgFtovYms-EFgtNliV6PyblqUUaRuI2yYdUny2k-am2dB2yVVnrgtM7htUFWKULC8u6NgwDgV9qm1nxlq8m900wqUiPiMp89248RgggOBtoB9AMQ_N1Xppp3s3eMokBTRN4urr1SZ-bjkQegTbnFZ9Zwv-TKUBKZ2Gd5VtbqZ2c5t62gauIJ2XcO3VkEjtsYzOP26fzbWmCfI8jzV0WDsxj0qEdKhzxVPLDERyvbN6VCwfiMUTLef1y_G1QJP3uYlkmz0ZvCQ",
+        "qi": "TgjJGFXrggHszRCRaccjeFJOHP22RMakqddAgImnMtStAFwOKdlNA1WND0xd4e1zVJso_IFRO9-kMrMv8JmLj-3QloS5-UFDlvfywfqHwPiyy3KVQtgGnm6PJ6WR2qbA2dcBgiGpM-lbC-t-mRa4YeNANmrTchr5RIlXNg4pVaFZUjG6-QDb4WL_CFmA5KgC70HFzaXFHvKFM2SYRo_lUytjINzsv6oayB6nFfhq0e1EyouyOY5cU5B57Lu9QSfMWdnbNdwW7pHNkZwOd9BqrRhBcO1JncHKsyr8bplWO1ffTftpMmSseDzaSd_Zi1n2h8dblu8jMh6pNAiWF46bBA",
+        "kid": "2011-04-29"
+    }
+    '
+
+    WALLET="$WALLET" run -0 bash -c 'echo "$WALLET" | yarn ardrive get-address -w /dev/stdin'
```
karlprieb, line 43: Thank you for your work 😄 
It's wonderful to have contributions like yours!

Sorry for being nitpicky :)
I don't think that we need to reassign `WALLET` with its own value.
It could be simple like that:
```suggestion
    run -0 bash -c 'echo $WALLET | yarn ardrive get-address -w /dev/stdin'
```

karlprieb (approved):

bats_test/get-address/get-address.bats
```diff
@@ -0,0 +1,46 @@
...
+        "dq": "B_WrEgZ8qY_3vLlJHsr5kJ26VqPgKiQRnJIsb6fsQdKOoUofTP3A0rzmZRAB1ltA-tChyBCMf9leBfgFtovYms-EFgtNliV6PyblqUUaRuI2yYdUny2k-am2dB2yVVnrgtM7htUFWKULC8u6NgwDgV9qm1nxlq8m900wqUiPiMp89248RgggOBtoB9AMQ_N1Xppp3s3eMokBTRN4urr1SZ-bjkQegTbnFZ9Zwv-TKUBKZ2Gd5VtbqZ2c5t62gauIJ2XcO3VkEjtsYzOP26fzbWmCfI8jzV0WDsxj0qEdKhzxVPLDERyvbN6VCwfiMUTLef1y_G1QJP3uYlkmz0ZvCQ",
+        "qi": "TgjJGFXrggHszRCRaccjeFJOHP22RMakqddAgImnMtStAFwOKdlNA1WND0xd4e1zVJso_IFRO9-kMrMv8JmLj-3QloS5-UFDlvfywfqHwPiyy3KVQtgGnm6PJ6WR2qbA2dcBgiGpM-lbC-t-mRa4YeNANmrTchr5RIlXNg4pVaFZUjG6-QDb4WL_CFmA5KgC70HFzaXFHvKFM2SYRo_lUytjINzsv6oayB6nFfhq0e1EyouyOY5cU5B57Lu9QSfMWdnbNdwW7pHNkZwOd9BqrRhBcO1JncHKsyr8bplWO1ffTftpMmSseDzaSd_Zi1n2h8dblu8jMh6pNAiWF46bBA",
+        "kid": "2011-04-29"
+    }
+    '
+
+    run -0 bash -c 'echo $WALLET | yarn ardrive get-address -w /dev/stdin'
```
matibat, line 43: Nice trick!

fedellen (approved):
🙌

xloem:
For me, the test failed if the WALLET environment variable is not exported to the subshell. Likely a cleaner approach now would be to simply remove the linebreaks between the variable definition and the run line. But if it works in you guys' test environment, that's great.

Thanks for all the work reviewing and addressing this PR.

karlprieb:
Hey @xloem how is the test failing?
We use https://github.com/ardriveapp/ardrive-bats-docker to run bats and it's working fine on my end.

About the multiline variable: we're using single quotes around the value ;)

xloem:
I'm not set up to test atm but could set up. `$WALLET` would evaluate to the empty string, since the variable wasn't exported. Why do you mention the single quotes?

PR #269, opened 2022-03-16
fix(duplicate pw prompt): Await each driveKey promise sequentially PE-1116

fedellen:
This PR fixes the password prompt to only prompt the user once per drive

### DIFF
--- a/src/commands/upload_file.ts
+++ b/src/commands/upload_file.ts
@@ -177,26 +177,33 @@ new CLICommand({
 				arweave
 			});
 
-			const uploadStats: ArDriveUploadStats[] = await Promise.all(
-				filesToUpload.map(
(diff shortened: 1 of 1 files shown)


src/commands/upload_file.ts
```diff
@@ -177,26 +177,33 @@ new CLICommand({
...
+					parentFolderId,
+					wrappedEntity,
+					destinationFileName,
+					driveKey,
+					drivePassword
+				} of filesToUpload) {
+					uploadStats.push({
+						wrappedEntity,
+						driveKey:
+							driveKey || (await parameters.getIsPrivate())
```
karlprieb, line 193: Let's use the nullish coalescing operator here, as it is safer than '||'.
```suggestion
							driveKey ?? (await parameters.getIsPrivate())
```

karlprieb (approved):

matibat (approved):
Excellent!

PR #271, opened 2022-03-24
PE-1148: Release ArDrive CLI v1.12.0

fedellen:
This PR merges `dev` to `master` and releases ArDrive CLI v1.12.0

Proposed GitHub Release Notes:

- Adds support for persistent caching of immutable transaction data to speed up drive operations
- Adds support for cache performance logging via the ARDRIVE_CACHE_LOG environment variable
- Fixes the private `-P` password prompt on `upload-file` when used with multiple files


matibat (approved):
The CORE dep version is 1.11.0 :+1:

PR #272, opened 2022-03-24
chore(version): Bump version to 1.13.0 for next release PE-1148

fedellen:


matibat (approved):
I don't yet know when do we have to bump also `main.dart?version=XX` at `web/index.html`

PR #289, opened 2022-06-08
PE-1682: Release ArDrive CLI v1.16.0

fedellen:
Proposed Notes:

- GraphQL requests are now more robust and will use a retry strategy
- Additional `stderr` logging added to all retried requests to an Arweave gateway
- Updates `arweave`, `arbundles`, and `smartweave` dependencies to avoid a catastrophic build error


matibat (approved):
Looks good. Version v1.16.0 ready to go!

PR #293, opened 2022-07-10
PE-877: Adds a --version flag to the CLI

arielmelendez:


bats_test/help/no_command.bats
```diff
@@ -33,23 +36,24 @@ Commands:
   rename-drive [options]
   rename-file [options]
   rename-folder [options]
+  retry-tx [options]
```
matibat, line 39: We missed to update the help when we added that feature

matibat (commented):
I've added `bats_require_minimum_version 1.5.0` to all of the files that uses BATS:run to suppress the warning:
```sh
BW02: Using flags on `run` requires at least BATS_VERSION=1.5.0. Use `bats_require_minimum_version 1.5.0` to fix this message.
```

Also, the code is formatted.

src/CLICommand/cli_command.ts
```diff
@@ -13,7 +16,14 @@ export interface CommandDescriptor {
...
 const programAsUnknown: unknown = program;
-const programApi: CliApiObject = programAsUnknown as CliApiObject;
+export const programApi: CliApiObject = programAsUnknown as CliApiObject;
+
+programApi.name('ardrive');
+programApi.version(CLI_APP_VERSION);
+programApi.addHelpCommand(true);
+programApi.usage('[command] [command-specific options]');
+// Override the commander's default exit (process.exit()) to avoid abruptly interrupting the script execution
+programApi.exitOverride();
```
fedellen, line 26: Why did these assignments get lifted up from the constructor into this module's scope?

fedellen (approved):

src/CLICommand/cli_command.ts
fedellen: You've answered here: https://github.com/ardriveapp/ardrive-cli/pull/293#issuecomment-1180862266 

:)

matibat: > I've made some changes to the CLICommand implementation because there was a node warning talking about a "possible listeners memory leak detected". https://stackoverflow.com/questions/9768444/possible-eventemitter-memory-leak-detected
> The listeners are now set only once. See: [2208aac](https://github.com/ardriveapp/ardrive-cli/pull/293/commits/2208aac4f144366cd396018b319edcfe36376a1c)

Because of a warning. The behavior is kept because the tests are running.

arielmelendez: This was actually how I originally coded it until the testing issue came up, at which point I moved everything back to the CLICommand constructor. This make much more logical sense - to set the program up 1 time.

PR #298, opened 2022-07-26
PE-1886: Integrate with Advanced Metadata (IPFS)

matibat:
Changes:
- Integrates the alpha version of CORE 
- Appends the IPFS CID to the custom metadata of each uploaded file


src/commands/upload_file.ts
```diff
@@ -121,6 +142,32 @@ function getSingleFile(parameters: ParametersHelper, parentFolderId: FolderID):
 	return [singleParameter];
 }
 
+async function getCustomMetaDataWithIpfsCid({
```
fedellen, line 145: Is it possible to handle this in the parameter helper?

fedellen (approved):
LGTM, left non-blocking question. Only because we can have some unit tests more easily from there

src/commands/upload_file.ts
fedellen: Following up on this, perhaps `getCustomMetaData` would check for the `--ipfs-add` flag internally

matibat: Moves the function to parameters helper: 3167dbc2ae8c54570ce919a11fd028dc53dce5f9
And adds unit tests: c703ac2ffb3f32cfe1be4c2a48a63a15b2fd6b8c

fedellen (approved):
Great updates!

PR #299, opened 2022-07-26
PE 1904/Implement remote path

jdaev:
Adds -remote-path flag to upload file command.


src/commands/upload_file.ts
```diff
@@ -159,6 +188,11 @@ new CLICommand({
 			if (fileList) {
 				return fileList;
 			}
+			const filesFromRemote = await getRemoteFile(parameters, parentFolderId);
+			console.log(filesFromRemote);
```
fedellen, line 192: extra `console.log`

src/utils/download.test.ts
```diff
@@ -0,0 +1,24 @@
...
+	});
+
+	it('download throws when given an invalid link', async () => {
+		let error;
+		try {
+			await download(invalidDownloadLink, tempFolderPath);
+		} catch (err) {
+			error = err;
+		}
+		expect(error?.name).to.equal('Error');
```
fedellen, line 22:
```suggestion
		expect(error?.name).to.equal('Error');
		expect(error?.message).to.equal('Invalid remote path. No file name found.');
```

src/utils/download.ts
```diff
@@ -0,0 +1,25 @@
...
+		throw new Error('Invalid remote path. No file name found.');
+	}
+	try {
+		const response: AxiosResponse = await axios.get(url, { responseType: 'arraybuffer' });
+		const pathToFile = path.join(filePath, fileName);
+
+		fs.writeFileSync(pathToFile, response.data);
+		return pathToFile;
+	} catch (error) {
+		throw new Error(`Failed to download file from remote path ${url}: ${error.message}`);
```
fedellen, line 23: non-blocking nit: We could gain test coverage over this error by exposing `axiosInstance` as an optional parameter and stub it during the tests with `stub(axiosInstance, 'get').resolves(throw Error('ERROR'))` (or similar, that may not be syntactically correct)

src/utils/temp.ts
```diff
@@ -0,0 +1,35 @@
...
+	return tempFolderPath;
+}
+
+export function cleanUpTempFolder(): void {
+	const tempFolderPath = platformTempFolder();
+	fs.readdir(tempFolderPath, (err, files) => {
+		if (err) throw err;
+		for (const file of files) {
+			fs.unlink(path.join(tempFolderPath, file), (err) => {
+				if (err) throw err;
```
fedellen, line 31: I don't mind this syntax, but as a standard we've always been using the brackets for more readable executions:

```suggestion
		if (err) { 
		    throw err;
		}
		for (const file of files) {
			fs.unlink(path.join(tempFolderPath, file), (err) => {
				if (err) {
				    throw err;
				}    
```

src/utils/temp.test.ts
```diff
@@ -0,0 +1,16 @@
+import { expect } from 'chai';
+import { getTempFolder } from './temp';
+import * as fs from 'fs';
+
+describe('temp folder test', () => {
```
fedellen, line 5: Non-blocking test nit:

We could improve this by adding or including a test for the cleanup function

```
desc: getTempFolder function returns a folder

let tempFolderPath
before: tempFolderPath = getTempFolder()

it: that exists (expect exist)
it: that contains the correct subfolders (expect folders)


desc: cleanUpTempFolder function

it: removes the temporary folder from the local system
  getTempFolder
  expect exist
  cleanUpFolder
  expect NOT exist
```

fedellen (approved):
LGTM

src/utils/download.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param url URL of the file to download.
+ * @param destinationPath Path to the destination file.
+ */
+export async function download(url: string, filePath: string): Promise<string> {
+	const fileName = url.split('/').pop();
+	if (!fileName) {
+		throw new Error('Invalid remote path. No file name found.');
+	}
+	try {
+		const response: AxiosResponse = await axios.get(url, { responseType: 'arraybuffer' });
```
arielmelendez, line 17: Could be nice to have some progress logging to stderr

arielmelendez: I don't think we have any good examples of progress logging from axios. But see `onUploadProgress` in https://axios-http.com/docs/req_config for inspo. Not a blocker but a nice to have if you can whip it up quickly.

arielmelendez: Can also see how we do naive progress logging for uploads in ArFSDao in ardrive-core-js for inspo. Search for `progressCallback` there.

src/utils/download.ts
```diff
@@ -0,0 +1,25 @@
+import * as fs from 'fs';
+import path from 'path';
+import axios, { AxiosResponse } from 'axios';
+
+/**
+ * Downloads file from remote HTTP[S] host and puts its contents to the
+ * specified location.
+ * @param url URL of the file to download.
+ * @param destinationPath Path to the destination file.
```
arielmelendez, line 9: I like this param name better, but it's `filePath` in your function (which I think is ambiguous).

src/utils/download.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export async function download(url: string, filePath: string): Promise<string> {
+	const fileName = url.split('/').pop();
+	if (!fileName) {
+		throw new Error('Invalid remote path. No file name found.');
```
arielmelendez, line 14: Why is obtaining a file name a requirement for this function? What if someone just wants to download the index page that gets returned from "https://whatever.com"? That seems valid.

I would have thought we would just generate a temporary filename in a tmp folder somewhere, and then have axios stream to that spot. Assuming that doesn't throw, then we use that tmp local filepath (whose name can be absolutely random... who cares?) to feed our normal upload code with whatever naming rules we want to apply THERE, rather than here. This should just be a utility downloader function. Filename derivation logic need not apply here.

src/utils/download.ts
```diff
@@ -0,0 +1,25 @@
...
+	}
+	try {
+		const response: AxiosResponse = await axios.get(url, { responseType: 'arraybuffer' });
+		const pathToFile = path.join(filePath, fileName);
+
+		fs.writeFileSync(pathToFile, response.data);
```
arielmelendez, line 20: As mentioned above, I'm worried about the scalability of this approach (i.e. reading the whole file into memory). I'd prefer if we streamed the data to disk. `pipeline` is a pretty clean way to do that that deals with a lot of error handling for you. See this SO: https://stackoverflow.com/a/64925465/1723537

src/commands/upload_file.ts
```diff
@@ -121,6 +123,32 @@ function getSingleFile(parameters: ParametersHelper, parentFolderId: FolderID):
...
+): Promise<UploadPathParameter[] | undefined> {
+	const remoteFilePath = parameters.getParameterValue(RemotePathParameter);
+
+	if (!remoteFilePath) {
+		return undefined;
+	}
+
+	const tempFolder = getTempFolder();
+
+	const localFilePath = await download(remoteFilePath, tempFolder);
```
arielmelendez, line 138: It occurs to me that we need some sensible requirements around choosing a file name when none is provided. Let's take to Slack as a team to hash something out. And really curious what @agsuy and @GiMartinezz have in mind.

arielmelendez (changes requested):
We need better requirements on upload filename choice when none is provided. Also need to stream the download to disk rather than loading into memory.

src/utils/download.ts
```diff
@@ -1,24 +1,46 @@
 import * as fs from 'fs';
 import path from 'path';
-import axios, { AxiosResponse } from 'axios';
+import axios from 'axios';
+import { uniqueId } from 'lodash';
```
fedellen, line 4: I think we can just do Number.random() or something native. We currently only include `lodash` because our seed phrase package would throw compilation errors

src/utils/download.ts
karlprieb: You can use `responseType: 'stream'`, get the length with `content-length` and get the current length with `on('data', (chunk) => chunk.length)`

Something like that:
```js
const {response, headers} = await axios.get(url, { responseType: 'stream' });
const totalLength = headers['content-length'];

const fileWriter = fs.createWriteStream(
  path.resolve(__dirname, fileName)
)

data.on('data', (chunk) => updateProgress(chunk.length))
data.pipe(fileWriter)
```

src/commands/upload_file.ts
```diff
@@ -121,6 +123,34 @@ function getSingleFile(parameters: ParametersHelper, parentFolderId: FolderID):
...
+	const remoteFilePath = parameters.getParameterValue(RemotePathParameter);
+
+	if (!remoteFilePath) {
+		return undefined;
+	}
+
+	const tempFolder = getTempFolder();
+
+	const localFilePath = await download(remoteFilePath, tempFolder, (downloadProgress: number) => {
+		console.error('Downloading file...', downloadProgress);
```
karlprieb, line 139: can we check for `ARDRIVE_PROGRESS_LOG` env variable to display progress?
Also, it would be good to use "process.stderr.write(\`Downloading file... ${downloadProgress}\r\`)" so it will replace the current terminal line and basically just use one line.

src/utils/download.ts
```diff
@@ -0,0 +1,55 @@
...
+				});
+				writer.on('close', () => {
+					if (!error) {
+						resolve(pathToFile);
+					}
+					//no need to call the reject here, as it will have been called in the
+					//'error' stream;
+				});
+			});
+		});
```
karlprieb, line 51: You could use `await` to clean up one level here.

```suggestion
		const response = await axios({
			method: 'get',
			url: url,
			responseType: 'stream',
			onDownloadProgress: downloadProgressCallback
		});

		return new Promise((resolve, reject) => {
			response.data.pipe(writer);
			let error: Error | null = null;

			writer.on('error', (err) => {
				error = err;

				writer.close();
				reject(err);
			});

			writer.on('close', () => {
				if (!error) {
					resolve(pathToFile);
				}
			});
		});
```

src/commands/upload_file.ts
```diff
@@ -121,6 +125,42 @@ function getSingleFile(parameters: ParametersHelper, parentFolderId: FolderID):
...
+
+	const tempFolder = getTempFolder();
+	const destinationFileName = parameters.getRequiredParameterValue(DestinationFileNameParameter);
+
+	const { pathToFile, contentType } = await download(
+		remoteFilePath,
+		tempFolder,
+		destinationFileName,
+		(downloadProgress: number) => {
+			if (process.env[ARDRIVE_PROGRESS_LOG] && process.env[ARDRIVE_PROGRESS_LOG] === '1') {
```
karlprieb, line 146: can we factor out this logic? Maybe create a `showProgressLog` const in utils file.

src/utils/download.ts
```diff
@@ -0,0 +1,53 @@
...
+type DownloadResult = { pathToFile: string; contentType: string };
+
+/**
+ * Downloads file from remote HTTP[S] host and puts its contents to the
+ * specified location.
+ * @param url URL of the file to download.
+ * @param destinationPath Path to the destination file.
+ * @param destinationFileName The file name.
+ */
+export async function download(
```
karlprieb, line 15: nit: can we add empty lines to separate the blocks and increase the legibility here?

src/commands/upload_file.ts
```diff
@@ -121,6 +124,42 @@ function getSingleFile(parameters: ParametersHelper, parentFolderId: FolderID):
...
+	const { pathToFile, contentType } = await download(
+		remoteFilePath,
+		tempFolder,
+		destinationFileName,
+		(downloadProgress: number) => {
+			if (showProgressLog) {
+				process.stderr.write(`Downloading file... ${downloadProgress}\r`);
+			}
+		}
+	);
```
fedellen, line 149: How often does this trigger, do we need a debounce involved here?

src/utils/download.ts
```diff
@@ -0,0 +1,54 @@
...
+			method: 'get',
+			url: url,
+			responseType: 'stream'
+		});
+		const totalLength = headers['content-length'];
+		const contentType = headers['content-type'];
+		let downloadedLength = 0;
+		data.on('data', (chunk: string | unknown[]) => {
+			downloadedLength += chunk.length;
+			downloadProgressCallback && downloadProgressCallback((downloadedLength / totalLength) * 100);
```
fedellen, line 35: nits:

```suggestion
			const downloadProgressPct = (downloadedLength / totalLength) * 100;
			downloadProgressCallback?.(downloadProgressPct);
```

src/utils/temp.ts
```diff
@@ -0,0 +1,28 @@
+import * as fs from 'fs';
+import * as os from 'os';
+import path from 'path';
+
+function platformTempFolder(): string {
```
fedellen, line 5: given that this file only handles temp folders, I would expect it to be named `temp_folder.ts` or similar. `temp.ts` is too generic

src/utils/download.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Downloads file from remote HTTP[S] host and puts its contents to the
+ * specified location.
+ * @param url URL of the file to download.
+ * @param destinationPath Path to the destination file.
+ * @param destinationFileName The file name.
+ */
+
+export async function download(
```
fedellen, line 16: Similar to the `temp` comment, `download` is too generic. It is clear here that we are only using this to download files. I would expect both the function and file to have a more descriptive name

fedellen (approved):
LGTM, Left some non blocking nits about file names and syntax

fedellen:
> LGTM, Left some non blocking nits about file names and syntax

And the debounce question, which may be important to get this through QA. We don't want to be spamming std.err obnoxiously

README.md
```diff
@@ -129,9 +129,10 @@ ardrive upload-file --wallet-file /path/to/my/wallet.json --parent-folder-id "f0
         15. [Uploading Manifests](#uploading-manifests)
         16. [Hosting a Webpage with Manifest](#hosting-a-webpage-with-manifest)
         17. [Uploading With a Custom Content Type](#custom-content-type)
-        18. [Uploading a Custom Manifest](#custom-manifest)
-        19. [Uploading Files with Custom MetaData](#uploading-files-with-custom-metadata)
-        20. [Applying Unique Custom MetaData During Bulk Workflows](#applying-unique-custom-metadata-during-bulk-workflows)
+        18. [Uploading From a Remote URL](#remote-path)
```
arielmelendez, line 132: Why this choice of location in the README? I think it might make more sense to have this one somewhere between items 4 and 5 🤔 I'll read the actual copy and think more about it.

README.md
```diff
@@ -1136,6 +1137,16 @@ It is currently possible to set this value to any given string, but the gateway
 
 Note: In the case of multi-file uploads or recursive folder uploads, setting this `--content-type` flag will set the provided custom content type on EVERY file entity within a given upload.
 
+### Uploading From a Remote URL<a id="remote-path"></a>
+
+You can upload a file from an existing url using the `--remote-path` flag. This must be used in conjunction with `--dest-file-name`
```
arielmelendez, line 1142: Nit: Missing ending punctuation

README.md
```diff
@@ -1136,6 +1137,16 @@ It is currently possible to set this value to any given string, but the gateway
...
+
+You can use a custom content type using the `--content-type` flag, but if this isn't used the app will use the content type from the response header:
```
arielmelendez, line 1144:
```suggestion
You can use a custom content type using the `--content-type` flag, but if this isn't used the app will use the content type from the response header of the request for the remote data:
```

src/commands/upload_file.ts
arielmelendez: From the looks of your demo video, very frequently. Could be worth debouncing it to achieve something like 3fps max. Non-blocking.

src/commands/upload_file.ts
```diff
@@ -134,12 +135,22 @@ async function getRemoteFile(
...
-	const localFilePath = await download(remoteFilePath, tempFolder);
+	const destinationFileName = parameters.getRequiredParameterValue(DestinationFileNameParameter);
+
+	const { pathToFile, contentType } = await download(
+		remoteFilePath,
+		tempFolder,
+		destinationFileName,
+		(downloadProgress: number) => {
+			if (showProgressLog) {
+				process.stderr.write(`Downloading file... ${downloadProgress}\r`);
```
arielmelendez:
```suggestion
				process.stderr.write(`Downloading file... ${downloadProgress.toFixed(1)}\r`);
```
This will keep it to 1 decimal place.

src/parameter_declarations.ts
```diff
@@ -284,8 +284,9 @@ Parameter.declare({
 Parameter.declare({
 	name: DestinationFileNameParameter,
 	aliases: ['-d', '--dest-file-name'],
-	description: `(OPTIONAL) a destination file name to use when uploaded to ArDrive
-\t\t\t\t\t\t\t• Only valid for use with --local-path or --local-file-path`
+	description: `a destination file name to use when uploaded to ArDrive
+\t\t\t\t\t\t\t• Required for use with --remote-path
+\t\t\t\t\t\t\t• Optional use with --local-path or --local-file-path`
```
arielmelendez, line 289:
```suggestion
\t\t\t\t\t\t\t• Optional when using with --local-path or --local-file-path`
```

src/utils/download.test.ts
```diff
@@ -6,16 +6,18 @@ import * as fs from 'fs';
 describe('download test', () => {
 	const validDownloadLink = 'https://arweave.net/pVoSqZgJUCiNw7oS6CtlVEd8gREQlpRbccrsMLkeIuQ';
 	const invalidDownloadLink = 'https://arweave.net/pVoSqZgJUCiNw7oS6CtlVEV8gREQlpRbccrsMLkeIuQ';
+	const name = 'cat.jpg';
```
arielmelendez: Nit: Overly general variable name. Consider `destinationFileName`, `stubDestFilename`, `testFileName`, etc.

src/utils/download.ts
arielmelendez: Going further, perhaps overkill but I never trust division of a variable:
```
const downloadProgressPct = (totalLength > 0) ? (downloadedLength / totalLength) * 100 : 0;
```

src/utils/download.ts
```diff
@@ -1,25 +1,54 @@
...
+		});
+		const totalLength = headers['content-length'];
+		const contentType = headers['content-type'];
+		let downloadedLength = 0;
+		data.on('data', (chunk: string | unknown[]) => {
+			downloadedLength += chunk.length;
+			downloadProgressCallback && downloadProgressCallback((downloadedLength / totalLength) * 100);
+		});
+		return new Promise((resolve) => {
+			data.pipe(writer);
```
arielmelendez: Were there problems using `pipeline` vs. `pipe`? I advocated strongly for the former due to the inherent difficulties in handling stream errors correctly. See here:
https://dev.to/morz/pipeline-api-the-best-way-to-handle-stream-errors-that-nobody-tells-you-about-122o

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,23 @@
+import { expect } from 'chai';
+import { cleanUpTempFolder, getTempFolder } from './temp_folder';
+import * as fs from 'fs';
+
+describe('temp folder test', () => {
```
arielmelendez, line 5:
```suggestion
describe('temp folder functions', () => {
```

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,23 @@
...
+	it('temp returns a folder that exists', () => {
```
arielmelendez, line 6: Add a nested describe here like:
```
describe('getTempFolder function', () => {
```
and put all the getTempFolder tests in it

arielmelendez: Then rename this test to
```
	it('returns a folder that exists', () => {
```
The net result of this is that the full test name is:
`temp folder functions getTempFolder function returns a folder that exists`

README.md
arielmelendez: Sure. I think we tried to build the ordering in a way that:
a) built upon concepts one after the other
b) optimized for most likely workflows sooner (as long as concepts from (a) weren't skipped)

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,23 @@
...
+		const tempFolderPath = getTempFolder();
+		expect(fs.existsSync(tempFolderPath)).to.equal(true);
+	});
+
+	it('temp path returned contains the correct subfolders', () => {
+		const tempFolderPath = getTempFolder();
+		expect(tempFolderPath).to.contains('ardrive');
```
arielmelendez, line 13: How about checking the os string in this test and doing a more precise expectation? e.g.
```
const expectedPathComponent = os.platform() === 'win32' ? 'ardrive-temp' : '.ardrive/temp';
expect(tempFolderPath).to.contains(expectedPathComponent);
```

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,23 @@
...
+		expect(tempFolderPath).to.contains('temp');
+	});
+
```
arielmelendez, line 16: Please add a test verifying that getTempFolder() can be called twice in a row without error.

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,23 @@
...
+	it('cleanUpTempFolder removes the temporary folder from the local system', () => {
+		const tempFolderPath = getTempFolder();
+		expect(fs.existsSync(tempFolderPath)).to.equal(true);
+		cleanUpTempFolder();
+		expect(fs.existsSync(tempFolderPath)).to.equal(false);
+	});
+});
```
arielmelendez, line 23: Please add 1 or more additional tests that verify that cleanupTempFolder can either (or both) be called twice without error or be called when there's nothing to clean up without error.

src/utils/download_file.ts
```diff
@@ -34,19 +39,8 @@ export async function downloadFile(
...
-				error = err;
-				writer.close();
-			});
-			writer.on('close', () => {
-				if (!error) {
-					resolve({ pathToFile, contentType });
-				}
-			});
-		});
+		await pipeline(data, writer);
```
arielmelendez, line 44: What behavior do you get right now if an error is thrown from one of the streams? e.g. if you pass a dead URL to try and download? Does pipeline throw?

arielmelendez: I only ask because we have no error handler here so I'm just wondering if the error ends up getting thrown (and whether that's what we want here).

src/commands/upload_file.ts
```diff
@@ -121,6 +124,42 @@ function getSingleFile(parameters: ParametersHelper, parentFolderId: FolderID):
...
+	const tempFolder = getTempFolder();
+	const destinationFileName = parameters.getRequiredParameterValue(DestinationFileNameParameter);
+
+	const { pathToFile, contentType } = await downloadFile(
+		remoteFilePath,
+		tempFolder,
+		destinationFileName,
+		(downloadProgress: number) => {
+			if (showProgressLog) {
+				process.stderr.write(`Downloading file... ${downloadProgress.toFixed(1)}% \r`);
```
arielmelendez, line 146: I'm getting an annoying console artifact due to the fact that we're using `write`:
```
yarn ardrive upload-file --remote-path "http://foo.com" --dest-file-name "wtvr.html"  -F 2d9386b7-25e5-40ea-a185-6a5508f8ed6a -w arweave-keyfile-9ODOd-_ZT9oWoRMVmmD4G5f9Z6MjvYxO3Nen-T5OXvU_cliSamplesmix.json --dry-run
{ownloading file... 0.0%
    "created": [

```
namely:
`{ownloading file... 0.0%`
You have 2 options:
1) Clear the line after the download is done:
```
    const { pathToFile, contentType } = await downloadFile(
		remoteFilePath,
		tempFolder,
		destinationFileName,
		(downloadProgress: number) => {
			if (showProgressLog) {
				process.stderr.write(`Downloading file... ${downloadProgress.toFixed(1)}% \r`);
			}
		}
	);
    process.stderr.clearLine(0);
```
2) Add a newline after the download, preserving the last downloading message:
```
    const { pathToFile, contentType } = await downloadFile(
		remoteFilePath,
		tempFolder,
		destinationFileName,
		(downloadProgress: number) => {
			if (showProgressLog) {
				process.stderr.write(`Downloading file... ${downloadProgress.toFixed(1)}% \r`);
			}
		}
	);
    process.stderr.write('\n');
```

arielmelendez (changes requested):
Left a few more suggestions, some requests for more tests, and a console log cleanup request.

src/commands/upload_file.ts
karlprieb: It is trigger for each downloaded chunk

src/commands/upload_file.ts
karlprieb: We should clear it so we have a clean output when tx are ready.

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,39 @@
+import { expect } from 'chai';
+import { cleanUpTempFolder, getTempFolder } from './temp_folder';
+import * as fs from 'fs';
+import * as os from 'os';
+
+describe('getTempFolder function', () => {
```
karlprieb, line 6: please use multiple describes:
- describe: "temp folder functions"
  - describe: "getTempFolder function"
    - it: 'returns a folder that exists'
    - it: 'can be called twice in a row'
    - it: 'returns a folder that contains the correct subfolders'
  - describe: "cleanUpTempFolder function"
    - it: 'removes the temporary folder from the local system'
    - it: 'can be called twice in a row'

src/commands/upload_file.ts
```diff
@@ -214,6 +258,7 @@ new CLICommand({
 			});
 
 			console.log(JSON.stringify(results, null, 4));
+			cleanUpTempFolder();
```
fedellen, line 261: Is this okay to cleanUp temp folder on each type of upload?

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,43 @@
...
+		});
+
+		it('returns a folder that contains the correct subfolders', () => {
+			const tempFolderPath = getTempFolder();
+			const expectedPathComponent = os.platform() === 'win32' ? '\\ardrive-temp' : '/.ardrive/temp';
+			expect(tempFolderPath).to.contains(expectedPathComponent);
+		});
+	});
+
+	describe('cleanUpTempFolder function', () => {
```
fedellen, line 27: What happens when cleanUpTempFolder is called when there is no tempFolder?

src/utils/download_file.test.ts
```diff
@@ -0,0 +1,29 @@
+import { expect } from 'chai';
+import { getTempFolder } from './temp_folder';
+import { downloadFile } from './download_file';
+import * as fs from 'fs';
+
+describe('downloadFile function', () => {
+	const validDownloadLink = 'https://arweave.net/pVoSqZgJUCiNw7oS6CtlVEd8gREQlpRbccrsMLkeIuQ';
+	const invalidDownloadLink = 'https://arweave.net/pVoSqZgJUCiNw7oS6CtlVEV8gREQlpRbccrsMLkeIuQ';
+	const destinationFileName = 'cat.jpg';
+	const tempFolderPath = getTempFolder();
```
fedellen, line 10: I would expect this folder to be cleaned up in `after(() => { cleanUpTempFolder() })`

fedellen (approved):

src/utils/temp_folder.test.ts
```diff
@@ -1,23 +1,43 @@
...
-		expect(fs.existsSync(tempFolderPath)).to.equal(true);
-	});
+describe('temp folder functions', () => {
+	describe('getTempFolder function', () => {
+		it('returns a folder that exists', () => {
+			const tempFolderPath = getTempFolder();
+			expect(fs.existsSync(tempFolderPath)).to.equal(true);
+		});
+
+		it('getTempFolder can be called twice in a row', () => {
```
arielmelendez, line 13:
```suggestion
		it('can be called twice in a row', () => {
```

src/utils/temp_folder.test.ts
```diff
@@ -1,23 +1,43 @@
...
+		});
 	});
 
-	it('cleanUpTempFolder removes the temporary folder from the local system', () => {
-		const tempFolderPath = getTempFolder();
-		expect(fs.existsSync(tempFolderPath)).to.equal(true);
-		cleanUpTempFolder();
-		expect(fs.existsSync(tempFolderPath)).to.equal(false);
+	describe('cleanUpTempFolder function', () => {
+		it('cleanUpTempFolder removes the temporary folder from the local system', () => {
```
arielmelendez, line 28:
```suggestion
		it('removes the temporary folder from the local system', () => {
```

src/utils/temp_folder.test.ts
```diff
@@ -1,23 +1,43 @@
...
+			const tempFolderPath = getTempFolder();
+			expect(fs.existsSync(tempFolderPath)).to.equal(true);
+			cleanUpTempFolder();
+			expect(fs.existsSync(tempFolderPath)).to.equal(false);
+		});
```
arielmelendez, line 33:
```suggestion
		});

```

src/utils/temp_folder.test.ts
```diff
@@ -1,23 +1,43 @@
...
+		it('cleanUpTempFolder can be called twice in a row', () => {
```
arielmelendez, line 34:
```suggestion
		it('can be called twice in a row', () => {
```

arielmelendez (approved):

src/commands/upload_file.ts
matibat: It is actually writing to STDERR. In my opinion the latter (writing the line break) is the best approach, but instead of that we should print the whole line again (because at the end of the printed string there is a carriage return).

_Not a blocker, I'm fine with the current approach as well_

So @arielmelendez, to avoid that behavior you should redirect the STDERR output to null:
```sh
$ yarn ardrive upload-file --remote-path "http://foo.com" --dest-file-name "wtvr.html"  -F 2d9386b7-25e5-40ea-a185-6a5508f8ed6a -w arweave-keyfile-9ODOd-_ZT9oWoRMVmmD4G5f9Z6MjvYxO3Nen-T5OXvU_cliSamplesmix.json --dry-run 2> /dev/null
```

src/parameter_declarations.ts
```diff
@@ -482,6 +490,24 @@ Parameter.declare({
...
+	name: RemotePathParameter,
+	aliases: ['--remote-path'],
+	description: `the remote path for the file that will be uploaded
+\t\t\t\t\t\t\t• MUST be used in conjunction with --dest-file-name
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-file-path
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-files
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-paths
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-path
+\t\t\t\t\t\t\t• Can NOT be used in conjunction with --local-csv`,
+	forbiddenConjunctionParameters: [
```
matibat, line 503: The Commander layer would handle the required parameters by itself
```suggestion
	requiredConjunctionParameters: [
		DestinationFileNameParameter
	],
	forbiddenConjunctionParameters: [
```

src/utils/temp_folder.test.ts
```diff
@@ -0,0 +1,44 @@
...
+		it('can be called twice in a row', () => {
+			const tempFolderPath = getTempFolder();
+			const tempFolderPath2 = getTempFolder();
+			expect(fs.existsSync(tempFolderPath)).to.equal(true);
+			expect(fs.existsSync(tempFolderPath2)).to.equal(true);
+		});
+
+		it('returns a folder that contains the correct subfolders', () => {
+			const tempFolderPath = getTempFolder();
+			const expectedPathComponent = os.platform() === 'win32' ? '\\ardrive-temp' : '/.ardrive/temp';
```
matibat, line 22: We usually avoid dynamic behaviors in the unit tests. In this case we are making use of `os.plarform` which is not a constant value. We could instead mock this all.

matibat (commented):
Left some comments

fedellen (approved):

src/utils/temp_folder.test.ts
arielmelendez: That's actually a pretty good idea. We could make the os string injectable and have it use os.platform as the default value if not specified. I would say non-blocking right now.

matibat (approved):
LGTM!

src/utils/temp_folder.test.ts
matibat: Not a blocker +1

arielmelendez (dismissed):

fedellen (dismissed):

src/utils/temp_folder.ts
```diff
@@ -2,11 +2,61 @@ import * as fs from 'fs';
...
+	return null;
+}
+
+function getValidUnixTempPath(): string | null {
+	const envTempFolder = process.env['TMPDIR'] ?? process.env['TMP'] ?? process.env['TEMP'];
+	if (envTempFolder) {
+		if (fs.existsSync(envTempFolder)) {
+			return envTempFolder;
+		} else {
+			const tempPath = path.join('tmp');
```
karlprieb, line 39: `tempPath` should be `/tmp` directly or `path.join('/tmp')`, but there's no reason to use `path.join` here.

karlprieb (approved):

karlprieb (approved):

PR #300, opened 2022-07-27
feat(custom metadata): Parse --metadata-json PE-1530

fedellen:


matibat (approved):
LGTM

README.md
```diff
@@ -1223,10 +1223,16 @@ This JSON schema object must contain instructions on where to put this metadata
 }
 ```
 
-Alternatively, the `--metadata-json` and `--metadata-gql-tags` parameters accept an array of string values to be applied to their respective target (MetaData Tx Data JSON or GQL Tags). This method of CLI input does not support multiple tag values for a given tag name and the input must be an EVEN number of string values. e.g:
+The `--metadata-gql-tags` parameter accepts an array of string values to be applied to the MetaData Tx GQL Tags. This method of CLI input does not support multiple tag values for a given tag name and the input must be an EVEN number of string values. e.g:
 
 ```shell
-upload-file --metadata-gql-tags "IPFS-Add" "MY_HASH" --metadata-json "Tag-1" "Tag-Val" "Tag-2" "Tag-Val" # ...
+upload-file --metadata-gql-tags "IPFS-Add" "MY_HASH" # ...
```
fedellen, line 1229:
```suggestion
upload-file --metadata-gql-tags "Custom Tag Name" "Custom Value" # ...
```

src/CLICommand/parameters_helper.test.ts
```diff
@@ -739,18 +739,15 @@ describe('ParametersHelper class', () => {
 				...baseArgv,
 				testCommandName,
 				'--metadata-json',
-				'JSON Field 1',
-				'Val 1',
-				'JSON Field 2',
-				'Val 2'
+				'{"key": "val", "key-2": true, "key-3": 420, "key-4": ["more", 1337]}'
```
arielmelendez, line 742: Looks good though I might also extend the test case to include nested objects.

arielmelendez (approved):

PR #301, opened 2022-07-27
feat(custom metadata): Excise --metadata-json PE-1530

fedellen:


README.md
```diff
@@ -1223,10 +1222,10 @@ This JSON schema object must contain instructions on where to put this metadata
 }
 ```
 
-Alternatively, the `--metadata-json` and `--metadata-gql-tags` parameters accept an array of string values to be applied to their respective target (MetaData Tx Data JSON or GQL Tags). This method of CLI input does not support multiple tag values for a given tag name and the input must be an EVEN number of string values. e.g:
+Alternatively, the `--metadata-gql-tags` parameters accepts an array of string values to be applied to the MetaData Tx GQL Tags. This method of CLI input does not support multiple tag values for a given tag name and the input must be an EVEN number of string values. e.g:
 
 ```shell
-upload-file --metadata-gql-tags "IPFS-Add" "MY_HASH" --metadata-json "Tag-1" "Tag-Val" "Tag-2" "Tag-Val" # ...
+upload-file --metadata-gql-tags "IPFS-Add" "MY_HASH"  # ...
```
matibat, line 1228: Nit:
I'd use a different example because it might be confusing: if someone tries that example they won't have the data but the metadata in IPFS

matibat (approved):
LGTM

README.md
fedellen, line 1228:
```suggestion
upload-file --metadata-gql-tags "Custom Tag Name" "Custom Tag Value"  # ...
```

arielmelendez:
Closing in favor of https://github.com/ardriveapp/ardrive-cli/pull/300

PR #310, opened 2022-08-09
PE 2037 Fix Upload Result sourceUrl

javdhu:
Checks remote parameter again and updated upload result.

### DIFF
--- a/src/commands/upload_file.ts
+++ b/src/commands/upload_file.ts
@@ -272,7 +272,7 @@ new CLICommand({
 				prompts: fileAndFolderUploadConflictPrompts
 			});
 
-			if (remoteFilePath && results.created[0]) {
+			if (remoteFilePath && results.created[0].type === 'file') {
 				// TODO: Include ArFSRemoteFileToUpload functionality in ArDrive Core
 				// TODO: Account for bulk remote path uploads in the future
 				results.created[0].sourceUri = remoteFilePath;


src/commands/upload_file.ts
```diff
@@ -227,6 +226,7 @@ new CLICommand({
 
 			const conflictResolution = parameters.getFileNameConflictResolution();
 			const shouldBundle = !!parameters.getParameterValue(ShouldBundleParameter);
+			const remoteFilePath = parameters.getParameterValue(RemotePathParameter);
```
fedellen, line 229: nit: Could improve the locality of this variable by declaring it about the if check

src/commands/upload_file.ts
```diff
@@ -272,6 +272,11 @@ new CLICommand({
 				prompts: fileAndFolderUploadConflictPrompts
 			});
 
+			if (remoteFilePath) {
+				// TODO: Include ArFSRemoteFileToUpload functionality in ArDrive Core
+				results.created[0].sourceUri = remoteFilePath;
+			}
+
 			console.log(JSON.stringify(results, null, 4));
 			cleanUpTempFolder();
```
fedellen, line 281: Would having `cleanUpTempFolder()` inside the new if check make sense?

fedellen (approved):

fedellen (approved):

arielmelendez (approved):

PR #316, opened 2022-08-12
Release v1.20.1

javdhu:
Release Notes:
- Prevent folder access errors for download cleanup on node versions 16 and above.


matibat (approved):
Ready to be shipped! 🚀

PR #319, opened 2022-08-15
PE-2075: use file stream while deriving ipfs hash

karlprieb:


matibat (commented):
It's possible to pass a stream here because `ReadStream` inherits from `Readable` the symbol `[Symbol.asyncIterator](): AsyncIterableIterator` which makes it of type `AsyncIterable<Uint8Array>` (sub-type of `CIDHashable`). See: [ReadStream](https://microsoft.github.io/PowerBI-JavaScript/classes/_node_modules__types_node_tty_d_._tty_.readstream.html)

LGTM - IDK if it's intended to point `master` here, so I'll approve after we confirm

matibat (approved):
+1

PR #322, opened 2022-08-31
fix(temp_folder): use rmSync instead of rmdirSync PE-2198

karlprieb:


src/utils/temp_folder.test.ts
```diff
@@ -19,8 +18,7 @@ describe('temp folder functions', () => {
 
 		it('returns a folder that contains the correct subfolders', () => {
 			const tempFolderPath = getTempFolder();
-			const expectedPathComponent =
-				os.platform() === 'win32' ? '\\ardrive-downloads' : '/.ardrive/ardrive-downloads';
+			const expectedPathComponent = 'ardrive-downloads';
```
matibat, line 21: Makes sense to me because it will be placed inside the tempFolderPath.

The test changed, but not the codebase? Is that because we didn't update the tests in a previous PR or because we are missing it in this PR?

matibat (commented):
One single question here

src/utils/temp_folder.test.ts
karlprieb: Exactly the there's no `.ardrive` folder when using `TMP`. That's why I changed the test to check only for `ardrive-downloads`.
The test would only pass for not `win32` if you don't have `TMP` or `TMPDIR` environment variable, so basically only on CI.

matibat (approved):
LGTM

PR #339, opened 2023-08-03
chore(deps): bump node, yarn, eslint, typescript, versions PE-1252

fedellen:


package.json
```diff
@@ -29,23 +29,23 @@
 		"@types/prompts": "^2.0.9",
 		"@types/sinon": "^10.0.2",
 		"@types/source-map-support": "^0",
-		"@typescript-eslint/eslint-plugin": "^4.18.0",
```
arielmelendez, line 32: @types/node above is still set to 14.x

arielmelendez: Do we still need node-fetch as an external dependency with Node 18?

src/CLICommand/cli_command.ts
```diff
@@ -62,7 +62,7 @@ function setCommanderCommand(commandDescriptor: CommandDescriptor, program: CliA
 				exitProgram(exitCode || 0);
 			});
 		} catch (e) {
-			console.error(`Error: ${e.message}`);
+			console.error(`Error: ${(e as Error).message}`);
```
arielmelendez, line 65: Might be safer to do a ternary check on the type and if it's not an Error then just string interpolate whatever bubbled out here.

src/CLICommand/parameters_helper.test.ts
```diff
@@ -610,7 +610,7 @@ describe('ParametersHelper class', () => {
 			}
 		});
 
-		it('throws on invalid URLs during the URL class constructor', async () => {
+		it.skip('throws on invalid URLs during the URL class constructor', async () => {
```
arielmelendez, line 613: What happened here?

src/utils/download_file.ts
```diff
@@ -45,6 +45,6 @@ export async function downloadFile(
 		return { pathToFile, contentType };
 	} catch (error) {
 		writer.close();
-		throw new Error(`Failed to download file from remote path ${url}: ${error.message}`);
+		throw new Error(`Failed to download file from remote path ${url}: ${(error as Error).message}`);
```
arielmelendez, line 48: Same comment about ternary type check and interpolation fallback.

arielmelendez (changes requested):
Highlighted some issues. Also be sure we don't need README updates.

package.json
fedellen: a5bbd1e

src/CLICommand/parameters_helper.test.ts
fedellen: The behavior of the `URL` error return seems to have changed and I'm getting some weird promise hangups in this test

src/CLICommand/cli_command.ts
fedellen: 5e860af

src/utils/download_file.ts
fedellen: 5e860af

fedellen:
> Highlighted some issues. Also be sure we don't need README updates.

Updated!

arielmelendez (approved):

PR #341, opened 2023-09-20
feat(turbo): use turbo sdk for upload PE-4472

fedellen:


src/CLICommand/parameters_helper.ts
```diff
@@ -466,7 +465,7 @@ export class ParametersHelper {
 
 		if (!userProvidedURL) {
 			// Return default CLI turbo if no turbo url can be derived from the user
-			return new URL(turboProdUrl);
+			return new URL('https://upload.ardrive.dev');
```
fedellen, line 468: TODO: should default to prod. but sending to dev temporarily

```suggestion
			return new URL('https://upload.ardrive.io');
```

PR #361, opened 2025-05-20
PE-8050: update for ArFS v0.15 and V2 drive signature-derived keys

kunstmusik:


src/commands/download_drive.ts
```diff
@@ -42,7 +42,9 @@ new CLICommand({
 				wallet,
 				arweave
 			});
-			const driveKey = await parameters.getDriveKey({ driveId });
+
+			const driveSignatureInfo = await ardrive.getDriveSignatureInfo(driveId, await wallet.getAddress());
+			const driveKey = await parameters.getDriveKey({ driveId, driveSignatureInfo });
```
arielmelendez, line 47: Small nit that this pattern is repeated so often it likely deserves its own utility function. One that either accepts parametershelper and an ardrive instance and a driveId or simply an expansion of the existing parametershelper method to accept an ardrive instance. Something to that effect.

src/commands/upload_file.ts
```diff
@@ -253,16 +253,23 @@ new CLICommand({
 					driveKey,
 					drivePassword
 				} of filesToUpload) {
+					let driveKeyForStats = driveKey;
+					if (!driveKeyForStats) {
+						const driveSignatureInfo = await arDrive.getDriveSignatureInfo(
```
arielmelendez, line 258: let's update the SDK function to take an object parameter with named keys so that users are less likely to pass a wallet address for a driveID and vice versa.

src/commands/download_drive.ts
kunstmusik: Updated in 2b69935aaa96b2212700ede94caab5594d756585

src/commands/upload_file.ts
kunstmusik: updated ardrive-core-js to add object parameter and updated code here in https://github.com/ardriveapp/ardrive-cli/pull/361/commits/5d83853d6f96ffa52f58d925c6ddab438b9d5289

